PR #1, opened 2024-02-26
Generate project structure using repo-config

llucax:
- **Generate project using repo-config**
- **Remove example files**
- **Add `part:client` GitHub label**
- **Fix the default branch for docs editing**
- **Sort and fix project's keywords**
- **Add project classifier**
- **Improve TOML formatting**
- **Remove TODOs that don't need addressing**
- **Add dependency to `grpcio`**


Marenz (approved):

.cookiecutter-replay.json
```diff
@@ -0,0 +1,46 @@
+{
+  "cookiecutter": {
+    "Introduction": "",
+    "type": "lib",
+    "name": "client-microgrid",
```
Marenz, line 5: Interesting, I used `frequenz-dispatch` here for the other repo. Following this logic I should have used `dispatch-actor`

leandro-lucarella-frequenz: No, just `dispatch` in this case. Not sure if you used the `actor` or `lib` type for it, but in both cases `dispatch` would have been the right choice, if it is a pure `actor` repo, it will build all the names using the `frequenz-actor-` prefix, for lib projects it just uses `frequenz-` and I think a `-python` suffix for the repo name. But this only affects the defaults, if you changed the other variables then it has no effect.

See https://frequenz-floss.github.io/frequenz-repo-config-python/v0.9/user-guide/start-a-new-project/#template-variables-reference.

llucax: No, just `dispatch` in this case. Not sure if you used the `actor` or `lib` type for it, but in both cases `dispatch` would have been the right choice, if it is a pure `actor` repo, it will build all the names using the `frequenz-actor-` prefix, for lib projects it just uses `frequenz-` and I think a `-python` suffix for the repo name. But this only affects the defaults, if you changed the other variables then it has no effect.

See https://frequenz-floss.github.io/frequenz-repo-config-python/v0.9/user-guide/start-a-new-project/#template-variables-reference (although I just realized there is no mention of how exactly other variables are built from this one).

Marenz: I did consult the reference before filling it out ;) 
Still wasn't clear if I should follow a certain pattern, e.g. we have every package prefixed with `frequenz-` etc

PR #9, opened 2024-02-27
Import code from the SDK v1.0.0-rc5

llucax:
The microgrid API client code is imported **as-is** from the SDK v1.0.0-rc5, except for some fixes to make `nox`/the CI pass:

- **Fix imports**
- **Add the missing `Fuse` type**
- **Workaround a weird `mypy` error about type mismatch in grpcio**
- **Fix wrong comparison**
- **Ignore pylint check**

Any other cleanups and improvements will be done in future PRs.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -4,4 +4,4 @@
 # Fallback owner.
 # These are the default owners for everything in the repo, unless a later match
 # takes precedence.
-* @frequenz-floss/api-microgrid-team
+* @frequenz-floss/api-microgrid-team @frequenz-floss/python-sdk-team
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-<!-- Here goes a general summary of what this release is about -->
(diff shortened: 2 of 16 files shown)


llucax:
@cwasicki @flora-hofmann-frequenz FYI (although is not really that interesting, because this is the code that's using the old microgrid API version, which is quite different from the new version, which is the one more similar to the reporting API).

Marenz (dismissed):
LGTM, though I am sad about the loss of Quantities for the fuse class

llucax:
> LGTM, though I am sad about the loss of Quantities for the fuse class

That should be addressed in the higher-level interface.

llucax:
I was going to merge so I can move forward, as it is just importing code, but I still need a review from a code owner (@frequenz-floss/api-microgrid-team). Not sure if we want to change this, at least for repos that need a fast pace during the initial development (either add more owners or remove the requirement that only code owners can really approve)

shsms:
Why is it owned by the microgrid api team?  They don't know this code and didn't write it either.  Unless you want to change the composition of the microgrid team,  this should be owned by the sdk team.

leandro-lucarella-frequenz:
I added a new commit to add the SDK team as an owner but we still need approval from the old owner to merge the change of ownership. I also added a discussion item for next meeting to talk about team structure.

leandro-lucarella-frequenz:
I'm force-merging to unblock this.

PR #11, opened 2024-02-28
Clean up the code

llucax:
- **Remove the MicrogridApiClient ABC**
- **Rename the class to `ApiClient`**
- **Unify the importing of protobuf generated files**
- **Don't disable the check `use-implicit-booleaness-not-comparison`**
- **Avoid disabling the `unused-argument` check for the whole file**
- **Import `grpc.aio` instead of `grpc`**
- **Improve `TypeVar` for `ComponentData`**
- **Fix import formatting**
- **Return `Self` in `from_proto()` constructors**
- **Remove nonexistent stub from the mock**
- **Add overrides to sub-classes**
- **Remove unnecessary import future**


tests/mock_api.py
```diff
@@ -152,8 +152,7 @@ def clear_bounds(self) -> None:
         """Drop all received bounds."""
         self._bounds.clear()
 
-    # pylint: disable=unused-argument
-    def ListComponents(
+    def ListComponents(  # pylint: disable=unused-argument
```
Marenz, line 155: Oooh, they are scoped?

Marenz (approved):

PR #12, opened 2024-02-28
Add more tests from the SDK v1.0.0-rc5

llucax:
- **Add more tests imported from the SDK v1.0.0-rc5**
- **Fix new tests imports**
- **Fix new tests patching**

### DIFF
--- a/tests/test_component.py
+++ b/tests/test_component.py
@@ -1,0 +1,94 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""Tests for the microgrid component wrapper."""
+
+import frequenz.api.common.components_pb2 as components_pb
+import frequenz.sdk.microgrid.component._component as cp
+import pytest
+
+# pylint:disable=no-member
+
+
+# pylint: disable=protected-access
+def test_component_category_from_protobuf() -> None:
+    """Test the creating component category from protobuf."""
+    assert (
+        cp._component_category_from_protobuf(
+            components_pb.ComponentCategory.COMPONENT_CATEGORY_UNSPECIFIED
+        )
+        == cp.ComponentCategory.NONE
+    )
+
+    assert (
+        cp._component_category_from_protobuf(
+            components_pb.ComponentCategory.COMPONENT_CATEGORY_GRID
+        )
+        == cp.ComponentCategory.GRID
+    )
+
+    assert (
+        cp._component_category_from_protobuf(
+            components_pb.ComponentCategory.COMPONENT_CATEGORY_METER
+        )
+        == cp.ComponentCategory.METER
+    )
+
+    assert (
+        cp._component_category_from_protobuf(
+            components_pb.ComponentCategory.COMPONENT_CATEGORY_INVERTER
+        )
+        == cp.ComponentCategory.INVERTER
+    )
+
+    assert (
+        cp._component_category_from_protobuf(
+            components_pb.ComponentCategory.COMPONENT_CATEGORY_BATTERY
+        )
+        == cp.ComponentCategory.BATTERY
+    )
+
+    assert (
+        cp._component_category_from_protobuf(
+            components_pb.ComponentCategory.COMPONENT_CATEGORY_EV_CHARGER
+        )
+        == cp.ComponentCategory.EV_CHARGER
+    )
+
+    assert cp._component_category_from_protobuf(666) == cp.ComponentCategory.NONE  # type: ignore
+
+    with pytest.raises(ValueError):
+        cp._component_category_from_protobuf(
+            components_pb.ComponentCategory.COMPONENT_CATEGORY_SENSOR
+        )
+
+
+# pylint: disable=invalid-name
+def test_Component() -> None:
+    """Test the component category."""
+    c0 = cp.Component(0, cp.ComponentCategory.GRID)
+    assert c0.is_valid()
+
+    c1 = cp.Component(1, cp.ComponentCategory.GRID)
+    assert c1.is_valid()
+
+    c4 = cp.Component(4, cp.ComponentCategory.METER)
+    assert c4.is_valid()
+
+    c5 = cp.Component(5, cp.ComponentCategory.INVERTER)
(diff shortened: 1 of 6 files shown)


llucax:
Based on #11, only the last 3 commits are relevant to this PR.

tests/test_component.py
```diff
@@ -0,0 +1,94 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
```
tiyash-basu-frequenz, line 2: The year should 2024 in all the new files.

tests/test_component.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+    assert (
+        cp._component_category_from_protobuf(
+            components_pb.ComponentCategory.COMPONENT_CATEGORY_EV_CHARGER
+        )
+        == cp.ComponentCategory.EV_CHARGER
+    )
+
+    assert cp._component_category_from_protobuf(666) == cp.ComponentCategory.NONE  # type: ignore
```
tiyash-basu-frequenz: Interesting.
`666` I assume is just a number that is not present in the `components_pb.ComponentCategory` enum? If yes, then getting a value like this would mean that something has gone horribly wrong, e.g., an API version mismatch.
Do we still want to return a `None` in this case, instead of an error?

tests/test_connection.py
```diff
@@ -0,0 +1,28 @@
...
+    assert not c10.is_valid()
+
+    c11 = client.Connection(1, 1)
+    assert not c11.is_valid()
+
+    c12 = client.Connection(1, 2)
+    assert c12.is_valid()
+
+    c21 = client.Connection(2, 1)
+    assert c21.is_valid()
```
tiyash-basu-frequenz, line 28: What would be an invalid connection? Shall we add a test for it?

tests/test_mock_api.py
```diff
@@ -0,0 +1,288 @@
...
+        Component(id=0, category=ComponentCategory.COMPONENT_CATEGORY_BATTERY),
+        Component(id=0, category=ComponentCategory.COMPONENT_CATEGORY_METER),
+    ]
+    assert (
+        list(api.ListConnections(ConnectionFilter(), service_context_mock).connections)
+        == []
+    )
+
+    # similarly, duplicates are allowed when adding new connections
+    api.add_connection(0, 0)
```
tiyash-basu-frequenz, line 76: I guess `(0, 0)` refer to the source and destination nodes? If yes, then there is a cycle in the component DAG, which makes it not-a-DAG. Why is this not an error case?

tests/test_timeout.py
```diff
@@ -0,0 +1,118 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""Benchmark for microgrid data."""
```
tiyash-basu-frequenz: Nitpick: Are these benchmarks, or just timing tests?

tests/test_component.py
leandro-lucarella-frequenz: I think we should preserve the original publication date, this file is copied from the SDK which was published in 2022.

llucax: I think we should preserve the original publication date, this file is copied from the SDK which was published in 2022.

tests/test_component.py
llucax: Good question. I'm not sure if @shsms wrote this originally and if he could remember if there was a particular reason for this.

In any case this is for a separate issue, this PR is just importing the code we have in the SDK right now. The idea is to get the same code here so we can do a first release that is exactly the same as the SDK (ok, there are some minor cleanup changes, but I wouldn't do any changes in semantics), so we can remove the code from the SDK and add this repo as a dependency and then start building here. Also much of the code here will radically change when we upgrade to the microgrid API v0.16, so I wouldn't spend much time fixing the present code.

I've been writing this in mostly every reply, so instead for every comment that I think should be addressed in a separate issue (and maybe even in the future, after the upgrade to v0.16) and sounding like a broken record, I will just add the :cd: emoji to indicate you should refer to this comment :laughing: 

But it would be definitely valuable to collect information about what's wrong with it so we don't make the same mistakes in the upgrade.

tests/test_connection.py
llucax: :cd: Yes, I think it is a good idea to add tests for this.

* https://github.com/frequenz-floss/frequenz-client-microgrid-python/issues/13

tests/test_component.py
tiyash-basu-frequenz: Fair enough. I'll suggest creating an issue to track this.

tests/test_mock_api.py
llucax: :cd: I will also defer this question to @shsms and wait for the answer to see if we need a separate issue for this.

tests/test_timeout.py
llucax: I'm 99% sure the comment is wrong, and this is just testing that if the server takes too long a timeout error will be raised. I will fix.

llucax: Fixed.

tests/test_component.py
llucax: OK, I will create discussions for the topics that are not clear we want to change, and issues for the ones that are obviously wrong and need fixing.

* https://github.com/frequenz-floss/frequenz-client-microgrid-python/discussions/14

tests/test_mock_api.py
llucax: * https://github.com/frequenz-floss/frequenz-client-microgrid-python/discussions/15

tests/test_component.py
shsms: I didn't write these, these are from before I started working on the SDK.

tiyash-basu-frequenz (approved):

tests/test_mock_api.py
shsms: agree here too, and it is from before my time on the SDK.

tests/test_timeout.py
shsms: This is likely me. \o/

PR #17, opened 2024-02-29
Use streaming helper and retry strategies from `frequenz-client-base`

llucax:
We depend on the git repo on a particular hash for `frequenz-client-base` to avoid having to do more upgrades than necessary in the short term, it will be hopefully replaced by a pypi release before the first release of this project.

- **Add missing dependency on `protobuf`**
- **Depend on `frequenz-client-base`**
- **docs: Show symbols type**
- **Use the `RetryStrategy` from `client-base`**
- **Remove retry strategies**
- **Use the `GrpcStreamingHelper` from `client-base`**


llucax:
Skipping release notes as this is an internal change only.

llucax:
I will do the v0.1.0 release after this is merged.

Marenz (approved):

PR #24, opened 2024-03-04
Downgrade protobuf to 4.21.6

llucax:
This version should work, as it is the version used in the SDK v1.0.0-rc5. If we don't do this we introduce an unnecessary dependency conflict with the SDK.


llucax:
Also pretty much trivial, so will auto-merge to make the v0.1.2 release.

PR #26, opened 2024-03-11
Update minimum `frequenz-channels` version to `v1.0.0-rc1`

shsms:


shsms:
Enabled auto-merge

Marenz (approved):

pyproject.toml
```diff
@@ -37,8 +37,8 @@ classifiers = [
 requires-python = ">= 3.11, < 4"
 dependencies = [
   "frequenz-api-microgrid >= 0.15.3, < 0.16.0",
-  "frequenz-channels == 1.0.0b2",
-  "frequenz-client-base >= 0.2.1, < 0.3.0",
+  "frequenz-channels == 1.0.0-rc1",
```
llucax, line 40: Too late now, but because we expect the final release to not have any further breaking changes, we could have used `>=` here instead of `==`, so we don't need to update it again once the final release is out.

llucax: * https://github.com/frequenz-floss/frequenz-client-microgrid-python/pull/31

PR #30, opened 2024-03-27
Add reactive power for inverters, meters and EV chargers

### LINKED ISSUE #28: Add reactive power for meters and inverters
### What's needed?

Meter and Inverter data types currently don't extract reactive power from the protobuf messages.  This needs to be supported.

llucax:
Fixes #28.

### DIFF
--- a/src/frequenz/client/microgrid/_component_data.py
+++ b/src/frequenz/client/microgrid/_component_data.py
@@ -87,6 +87,22 @@ class MeterData(ComponentData):
     * Negative means supply into the grid.
     """
 
+    reactive_power: float
+    """The total reactive 3-phase AC power, in Volt-Ampere Reactive (VAr).
+
+    * Positive power means capacitive (current leading w.r.t. voltage).
+    * Negative power means inductive (current lagging w.r.t. voltage).
+    """
+
+    reactive_power_per_phase: tuple[float, float, float]
+    """The per-phase AC reactive power, in Volt-Ampere Reactive (VAr).
+
+    The provided values are for phase 1, 2, and 3 respectively.
+
+    * Positive power means capacitive (current leading w.r.t. voltage).
+    * Negative power means inductive (current lagging w.r.t. voltage).
+    """
+
     current_per_phase: tuple[float, float, float]
     """AC current in Amperes (A) for phase/line 1,2 and 3 respectively.
 
@@ -122,6 +138,12 @@ class MeterData(ComponentData):
                 raw.meter.data.ac.phase_1.power_active.value,
                 raw.meter.data.ac.phase_2.power_active.value,
                 raw.meter.data.ac.phase_3.power_active.value,
+            ),
+            reactive_power=raw.meter.data.ac.power_reactive.value,
+            reactive_power_per_phase=(
+                raw.meter.data.ac.phase_1.power_reactive.value,
+                raw.meter.data.ac.phase_2.power_reactive.value,
+                raw.meter.data.ac.phase_3.power_reactive.value,
             ),
             current_per_phase=(
                 raw.meter.data.ac.phase_1.current.value,
@@ -270,6 +292,22 @@ class InverterData(ComponentData):  # pylint: disable=too-many-instance-attribut
     * Negative means supply into the grid.
     """
 
+    reactive_power: float
+    """The total reactive 3-phase AC power, in Volt-Ampere Reactive (VAr).
+
+    * Positive power means capacitive (current leading w.r.t. voltage).
+    * Negative power means inductive (current lagging w.r.t. voltage).
+    """
+
+    reactive_power_per_phase: tuple[float, float, float]
+    """The per-phase AC reactive power, in Volt-Ampere Reactive (VAr).
+
+    The provided values are for phase 1, 2, and 3 respectively.
+
+    * Positive power means capacitive (current leading w.r.t. voltage).
+    * Negative power means inductive (current lagging w.r.t. voltage).
+    """
+
     current_per_phase: tuple[float, float, float]
     """AC current in Amperes (A) for phase/line 1, 2 and 3 respectively.
 
@@ -359,6 +397,12 @@ class InverterData(ComponentData):  # pylint: disable=too-many-instance-attribut
                 raw.inverter.data.ac.phase_2.power_active.value,
                 raw.inverter.data.ac.phase_3.power_active.value,
             ),
+            reactive_power=raw.inverter.data.ac.power_reactive.value,
+            reactive_power_per_phase=(
+                raw.inverter.data.ac.phase_1.power_reactive.value,
+                raw.inverter.data.ac.phase_2.power_reactive.value,
+                raw.inverter.data.ac.phase_3.power_reactive.value,
+            ),
             current_per_phase=(
                 raw.inverter.data.ac.phase_1.current.value,
                 raw.inverter.data.ac.phase_2.current.value,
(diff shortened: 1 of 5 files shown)


matthias-wende-frequenz (commented):
Adding the metric to the `ComponentMetricId` enum is missing.
https://github.com/frequenz-floss/frequenz-client-microgrid-python/blob/v0.x.x/src/frequenz/client/microgrid/_component.py#L197

llucax:
Fixed and added release notes. I will do a release after it is merged.

src/frequenz/client/microgrid/_component_data.py
```diff
@@ -70,18 +70,30 @@ class MeterData(ComponentData):
     """A wrapper class for holding meter data."""
 
     active_power: float
-    """The 3-phase active power, in Watts, represented in the passive sign convention.
-            +ve current means consumption, away from the grid.
-            -ve current means supply into the grid.
+    """The 3-phase active power, in W (Watt).
```
tiyash-basu-frequenz: Maybe it's better to be explicit here?
```suggestion
    """The total 3-phase active power, in W (Watt).
```

src/frequenz/client/microgrid/_component_data.py
```diff
@@ -363,18 +387,30 @@ class EVChargerData(ComponentData):  # pylint: disable=too-many-instance-attribu
     """A wrapper class for holding ev_charger data."""
 
     active_power: float
-    """The 3-phase active power, in Watts, represented in the passive sign convention.
-        +ve current means consumption, away from the grid.
-        -ve current means supply into the grid.
+    """The 3-phase active power, in Watts (W).
```
tiyash-basu-frequenz: Maybe it's better to be explicit here?
```suggestion
    """The total 3-phase active power, in W (Watt).
```

src/frequenz/client/microgrid/_component_data.py
```diff
@@ -87,6 +87,24 @@ class MeterData(ComponentData):
     * `-` means supply into the grid.
     """
 
+    reactive_power: float
+    """The 3-phase reactive power, in Volt-Ampere Reactive (VAr).
```
tiyash-basu-frequenz: Maybe it's better to be explicit here?
```suggestion
    """The total 3-phase reactive power, in Volt-Ampere Reactive (VAr).
```

src/frequenz/client/microgrid/_component_data.py
```diff
@@ -87,6 +87,24 @@ class MeterData(ComponentData):
...
+
+    Represented in the passive sign convention.
+
+    * `+` means consumption from the grid.
+    * `-` means supply into the grid.
```
shsms: I think the terminology here is Capacitive/Inductive.  My guess is that it is not dependent on the grid.  And not passive sign convention.

src/frequenz/client/microgrid/_component_data.py
```diff
@@ -87,6 +87,24 @@ class MeterData(ComponentData):
...
+    """
```
tiyash-basu-frequenz: Active/passive sign convention does not apply to reactive power. So this needs to be changed (here and everywhere else).

This is what we do currently: https://github.com/frequenz-floss/frequenz-api-microgrid/blob/2332734e86dd8c1d3bf4ee601b93d52525654a6f/proto/frequenz/api/microgrid/microgrid.proto#L390-L393
.. and we plan to flip it in v0.17.0: https://github.com/frequenz-floss/frequenz-api-microgrid/issues/222

src/frequenz/client/microgrid/_component_data.py
matthias-wende-frequenz: True. Alternatively leading or lagging.

src/frequenz/client/microgrid/_component_data.py
llucax: Okay, this is definitely out of my expertise, so I guess I will just copy&paste what the API docs say. BTW, what is `ve` in those comments? I thought it was volt-electron but apparently that doesn't exist (I guess I was mixing it up with `eV`), and that it was a copy&paste error, but it seems it is neither, so I wonder what is it.

src/frequenz/client/microgrid/_component_data.py
tiyash-basu-frequenz: We use Leading/Lagging in the API, so better to use that.

src/frequenz/client/microgrid/_component_data.py
tiyash-basu-frequenz: > I guess I will just copy&paste what the API docs say.

Yeah, sounds good. We need to flip it here when we move this to API v0.17.0.

> BTW, what is ve in those comments?

`-ve` => negative, and `+ve` => positive :)
Probably I'll change it to `negative` and `positive` given this confused you.

llucax: Haha, yeah, totally unexpected, I never saw that convention before.

llucax: There seems to be an inconsistency, one says positive is capacitive (leading) and the other say the opposite (positive is inductive (lagging)):

* https://github.com/frequenz-floss/frequenz-api-common/blob/08538ce91ebec107668628c14498fa0ae00253ad/proto/frequenz/api/common/metrics/electrical.proto#L108-L112
  ```protobuf
    // AC reactive power.
    // +ve power means inductive (leading).
    // -ve power means capacitive (lagging).
    // In Volt-Ampere reactive (VAr).
    metrics.Metric power_reactive = 4;
  ```

* https://github.com/frequenz-floss/frequenz-api-common/blob/08538ce91ebec107668628c14498fa0ae00253ad/proto/frequenz/api/common/metrics/electrical.proto#L161-L165

  ```protobuf
  // The reactive 3-phase AC power.
  // +ve power means capacitive (current leading w.r.t. voltage).
  // -ve power means inductive (current lagging w.r.t. voltage).
  // In Volt-Ampere reactive (VAr).
  metrics.Metric power_reactive = 5;
  ```

Or is this correct? If it is not, which one is correct?

While trying to see if the inconsistency was also present in the current version, I noticed that we *lost* all documentation about what each metric represents, the unit etc. Is this correct or am I missing anything.

As far as I see now key are using a key-value approach, where the key is the metric ID and the value is either a float or an aggregated value (with max/min/avg/etc). I guess the documentation about each metric should now be in the `enum Metric` values, but there is pretty much nothing there.

If this is correct, I think we should re-add the documentation for each metric, it is a very valuable piece of information we are losing otherwise. :scream: 

One downside about the current approach compared to the old one, is we now need to duplicate the documentation (or come up with a clever way to cross-reference some central documentation) for metrics that are basically the same, like active/reactive/apparent power for each phase.

llucax: Judging from https://github.com/frequenz-floss/frequenz-api-microgrid/issues/222 I assume negative = inductive is correct. So I updated the PR with this in mind.

tiyash-basu-frequenz: >There seems to be an inconsistency, one says positive is capacitive (leading) and the other say the opposite (positive is inductive (lagging)):

Thanks for pointing out. It looks like a documentastion error.
https://github.com/frequenz-floss/frequenz-api-common/blob/08538ce91ebec107668628c14498fa0ae00253ad/proto/frequenz/api/common/metrics/electrical.proto#L108-L112
This is wrong^ . Inductive power should be Lagging (current lags voltage).

tiyash-basu-frequenz: >I assume negative = inductive is correct.

According to our present standards, yes. Once https://github.com/frequenz-floss/frequenz-api-microgrid/issues/222 is addressed, it will need to be flipped.

tiyash-basu-frequenz (approved):
LGTM!

PR #37, opened 2024-04-11
Convert to `betterproto`

llucax:
This PR migrated the Microgrid API to use `betterproto` instead of `grpc`.

A couple of bugs were found on the way and fixed, one of them directly related to the lack of proper type hints in `grpc`.

The changes are fairly small.

> [!NOTE]
> This PR is just a draft, tests need to be adapted still and I don't plan to merge this to `v0.x.x` directly, as it would be an unnecessary breaking change for the branch used by the SDK. This is mainly to show what changes are needed to migrate a project to `betterproto`.

This PR is using an unreleased repository that hosts the files generated by betterproto from the protobuf files. It will be published soon :tm:


pyproject.toml
```diff
@@ -36,9 +32,10 @@ classifiers = [
 ]
 requires-python = ">= 3.11, < 4"
 dependencies = [
-  "frequenz-api-microgrid >= 0.15.3, < 0.16.0",
+  "betterproto == 2.0.0b6",
   "frequenz-channels >= 1.0.0-rc1, < 2.0.0",
   "frequenz-client-base >= 0.3.0, < 0.4.0",
+  "frequenz-microgrid-betterproto @ file:///home/luca/devel/frequenz-microgrid-betterproto",
```
cwasicki: Is this intended?

cwasicki: >an unreleased repository

Is this meant here?

cwasicki (commented):
Wouldn't the first two commits make sense even when we not migrate to betterproto?

pyproject.toml
llucax: Yes, this is just a draft, I didn't crated those repos yet. Will do so soon.

llucax:
> Wouldn't the first two commits make sense even when we not migrate to betterproto?

Yes, also fixing the missing `await`, I just left it like that for "presentation purposes", but I think it makes sense to make a separate PR fixing the other issues in this PR that are not related to betterproto :+1:

llucax:
This will not pass the tests because it is still waiting for a release of microgrid-betterproto, but it should be ready for the final review, as no code changes should be needed.

llucax:
Now everything is green :green_heart:

llucax:
Enabling auto-merge.

pyproject.toml
```diff
@@ -2,11 +2,7 @@
 # Copyright © 2024 Frequenz Energy-as-a-Service GmbH
 
 [build-system]
-requires = [
-  "setuptools == 68.1.0",
-  "setuptools_scm[toml] == 7.1.0",
-  "frequenz-repo-config[lib] == 0.9.1",
```
cwasicki, line 8: Why is repo-config removed here?

llucax: Including `repo-config` as a build dependency in every project was actually a mistake, it was only needed for API projects and now that this is taken care by `setuptools-betterproto` (if we do the switch), it will not be used by any project. My idea is to split repo-config even further in the future, it was an experiment and now that we have a much clear picture about what we need from it, it is a good time to split it into smaller projects and use those smaller projects only where necessary.

llucax: It is unrelated to this PR, though, I just did it now because I realized it was a mistake to include `repo-config` in the build dependencies right after creating `setuptools-betterproto`.

src/frequenz/client/microgrid/_component_data.py
```diff
@@ -390,7 +382,7 @@ def from_proto(cls, raw: PbComponentData) -> Self:
         raw_power = raw.inverter.data.ac.power_active
         inverter_data = cls(
             component_id=raw.id,
-            timestamp=raw.ts.ToDatetime(tzinfo=timezone.utc),
```
Marenz, line 393: Hmm we have functions for the conversion in the client library

Marenz (approved):

src/frequenz/client/microgrid/_component_data.py
llucax: This comes from the SDK, predates all the recent work we've been doing with APIs.

PR #42, opened 2024-05-15
Fix a few bugs related to `GrpcStreamBroadcaster`

llucax:
When a broadcaster already existed, a new one was created but immediately discarded, as it was never saved. Instead we really want to avoid creating a new broadcaster if one already exists.

Also when introducing the `GrpcStreamBroadcaster` from `client-base` we forgot to forward the selected retry strategy to it.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,4 +14,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+- Fix a leakage of `GrpcStreamBroadcaster` instances.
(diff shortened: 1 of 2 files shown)


llucax:
Enabled auto-merge.

daniel-zullo-frequenz (approved):
`dict.setdefault()` is sneakily hiding the leakage, good catch! LGTM

llucax:
> `dict.setdefault()` is sneakily hiding the leakage, good catch! LGTM

Yeah, I guess that's one disadvantage of using `dict.setdefault()`, it is not super obvious what's going on.

PR #46, opened 2024-05-15
Allow using `None` to specify the default retry strategy

llucax:
This is very useful when forwarding arguments.

### DIFF
--- a/src/frequenz/client/microgrid/_client.py
+++ b/src/frequenz/client/microgrid/_client.py
@@ -67,7 +67,7 @@ class ApiClient:
         self,
         grpc_channel: grpc.aio.Channel,
         target: str,
-        retry_strategy: retry.Strategy = retry.LinearBackoff(),
+        retry_strategy: retry.Strategy | None = None,
     ) -> None:
         """Initialize the class instance.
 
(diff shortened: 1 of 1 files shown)


src/frequenz/client/microgrid/_client.py
```diff
@@ -67,7 +67,7 @@ def __init__(
         self,
         grpc_channel: grpc.aio.Channel,
         target: str,
-        retry_strategy: retry.Strategy = retry.LinearBackoff(),
+        retry_strategy: retry.Strategy | None = None,
```
tiyash-basu-frequenz, line 70: The default being specified here is `None`, but the docs suggest that the default has to be linear backoff. Is this how it is supposed to work?

llucax: Good point, it is actually leaking the implementation details. We use another class to handle the data streams, that take the retry strategy as an argument, and also accepts `None` and uses a linear backoff by default when `None` is specified.

Not sure here what's best, from the user perspective I think is best to know exactly what happens with the default, but from the development POV it could be harder to maintain as if the default changes in the underlying class, then the docs have the risk of getting inadvertently outdated. :thinking:

llucax: BTW, the underlying class (`GrpcStreamBroadcaster`) is definitely an implementation detail, and shouldn't be exposed to the library users, so it is not an option to say "for the default go check this internal class").

Marenz (approved):

PR #47, opened 2024-05-15
Properly await for `AddInclusionBounds`

llucax:
Without this, the function call will never actually run. I'm not sure if this was working anyways because of some obscure implementation detail in grpcio, but in normal python a not awaited async function should never actually run.

Besides this, for sure errors can't be really caught without the await either.

This was detected while migrating to betterproto, as betterproto has correct type hints.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,3 +16,4 @@
 
 - Fix a leakage of `GrpcStreamBroadcaster` instances.
 - The user-passed retry strategy was not properly used by the data streaming methods.
+- The client `set_bounds()` method might have not done anything and if it did, errors were not properly raised.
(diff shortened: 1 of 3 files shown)


shsms:
Btw, the SDK doesn't use the `set_bounds` method anymore.  We use `set_power` for both EV chargers and PV inverters.

The previous version of the EV charger pool was using this implementation of `set_bounds`, but that was never deployed.

The older version of the EV charger that was deployed in a demo installation was using an older streaming `set_bounds` method,  so this existing one was never used, and won't be used anymore by the SDK.

llucax:
OK, I guess that explains why we didn't notice.

shsms (approved):

Issue #45, opened 2024-05-15
Add a specific client errors


llucax:
### What's needed?

Client errors should be easy to catch in a pythonic way, so we should provide one separate error per gRPC status code, at least.

### Proposed solution

Add `ClientTimeout`, `RequestCancelled`, etc. exceptions inheriting from `ClientError`.

### Additional context

* https://grpc.io/docs/guides/error/

PR #48, opened 2024-05-16
Raise a new `ClientError` instead of `AioRpcError`

llucax:
We want to avoid exposing the underlying details of the client, so in the future we can change the libraries used to connect to the service, or even the protocol.

This also removes the re-building of the exceptions, which was pretty unnecessary and needed some `type: ignore`s due to typing issues in `grpcio`.

We should probably create a separate error for timeouts in the future.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,8 @@
 
 ## Upgrading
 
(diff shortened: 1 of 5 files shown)


llucax:
BTW, I plan to add more error classes, see #45, but I will do it after migrating to betterproto/grpclib, so I don't need to adapt all the new error classes to grpclib twice.

tiyash-basu-frequenz (approved):

PR #49, opened 2024-05-24
Make the client accept a server URL to connect to

### LINKED ISSUE #50: Use a connection string instead of a `grpclib.client.Channel`

llucax:
This URL is a string with the form `grpc://hostname[:<port:int=9090>][?ssl=<ssl:bool=false>]`, meaning that the `port` and `ssl` are optional and default to 9090 and `false` respectively.

This makes it more convenient to take server URLs from config files and enviroment variables, but it also means gRPC channels are not exposed directly to users anymore, so the internal implementation to connect to the gRPC server can be changed without it being a breaking change (we could potentially even change the protocol completely to use something else rather than gRPC).

### DIFF
--- a/src/frequenz/client/microgrid/_client.py
+++ b/src/frequenz/client/microgrid/_client.py
@@ -12,7 +12,7 @@ import grpclib
 import grpclib.client
 from betterproto.lib.google import protobuf as pb_google
 from frequenz.channels import Receiver
-from frequenz.client.base import retry, streaming
+from frequenz.client.base import channel, retry, streaming
 from frequenz.microgrid.betterproto.frequenz.api import microgrid as pb_microgrid
 from frequenz.microgrid.betterproto.frequenz.api.common import (
     components as pb_components,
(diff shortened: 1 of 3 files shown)


llucax:
Enabling auto-merge.

src/frequenz/client/microgrid/_client.py
```diff
@@ -48,33 +48,41 @@
...
 class ApiClient:
-    """Microgrid API client implementation using gRPC as the underlying protocol."""
+    """A microgrid API client."""
 
     def __init__(
         self,
-        grpc_channel: grpclib.client.Channel,
-        target: str,
+        server_url: str,
+        *,
```
daniel-zullo-frequenz, line 56: Should we need to explicitly clarify in the release-notes that the `ApiClient` no longer needs the `grpc_channel` parameter to be constructed?

daniel-zullo-frequenz (commented):
I have a question and LGTM

src/frequenz/client/microgrid/_client.py
llucax: Not sure what do you mean specifically, the release notes currently say:

- The client now uses a string URL to connect to the server. The current accepted format is `grpc://hostname[:<port:int=9090>][?ssl=<ssl:bool=false>]`, meaning that the `port` and `ssl` are optional and default to 9090 and `false` respectively. You will have to adapt the way you connect to the server in your code.

Isn't that saying what you are asking for? You mean saying explicitly that the argument `target` and `grpc_channel` were merged into `server_url`?

daniel-zullo-frequenz: >  Isn't that saying what you are asking for? You mean saying explicitly that the argument target and grpc_channel were merged into server_url?

Yes, somehow it is. But I think it's useful for the user to make it explicit as as you just suggested it

llucax: Updated.

daniel-zullo-frequenz (approved):

PR #54, opened 2024-05-29
Add missing state and error wrappers

### LINKED ISSUE #40: Wrap missing state and errors
### What's needed?

The SDK is using some symbols directly from the protobuf generated files.

### Proposed solution

Wrap it so it can be used from the client instead.

### Use cases

* https://github.com/frequenz-floss/frequenz-sdk-python/blob/a31c84506b1a15761308ceb5c982a99d0c3f9948/src/frequenz/sdk/actor/power_distributing/_component_status/_pv_inverter_status_tracker.py#L14-L16	
* https://github.com/frequenz-floss/frequenz-sdk-python/blob/a31c84506b1a15761308ceb5c982a99d0c3f9948/src/frequenz/sdk/actor/power_distributing/_component_status/_battery_status_tracker.py#L25-L28
* Maybe more

llucax:
- **Remove some unnecessary pylint exceptions**
- **Add a missing `UNKNOWN` EV charger component state**
- **Return `UNKNOWN` instead of `UNSPECIFIED` for unknown states**
- **Wrap missing component errors**
- **Wrap missing component states**
- **Unify the style with the new wrappers**
- **Update release notes**
- **Prepare release notes for the v0.4.0 release**

### DIFF
--- a/src/frequenz/client/microgrid/_component_error.py
+++ b/src/frequenz/client/microgrid/_component_error.py
@@ -1,0 +1,213 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Common definitions for the microgrid API."""
+
+from dataclasses import dataclass
+from enum import Enum
+from typing import Self
+
+from frequenz.microgrid.betterproto.frequenz.api.microgrid import (
+    battery,
+    common,
+    inverter,
+)
+
+
+class ErrorLevel(Enum):
+    """Error level."""
+
+    UNSPECIFIED = common.ErrorLevel.ERROR_LEVEL_UNSPECIFIED
+    """Unspecified component error."""
+
+    WARN = common.ErrorLevel.ERROR_LEVEL_WARN
+    """Action must be taken to prevent a severe error from occurring in the future."""
+
+    CRITICAL = common.ErrorLevel.ERROR_LEVEL_CRITICAL
+    """A severe error that causes the component to fail. Immediate action must be taken."""
+
+    @classmethod
+    def from_pb(cls, code: common.ErrorLevel) -> Self:
+        """Convert a protobuf error level value to this enum.
+
+        Args:
+            code: The protobuf error level to convert.
+
+        Returns:
+            The enum value corresponding to the protobuf message.
+        """
+        try:
+            return cls(code)
+        except ValueError:
+            return cls(cls.UNSPECIFIED)
+
+
+class BatteryErrorCode(Enum):
+    """Battery error code."""
+
+    UNSPECIFIED = battery.ErrorCode.ERROR_CODE_UNSPECIFIED
+    """Unspecified battery error code."""
+
+    HIGH_CURRENT_CHARGE = battery.ErrorCode.ERROR_CODE_HIGH_CURRENT_CHARGE
+    """Charge current is too high."""
+
+    HIGH_CURRENT_DISCHARGE = battery.ErrorCode.ERROR_CODE_HIGH_CURRENT_DISCHARGE
+    """Discharge current is too high."""
+
+    HIGH_VOLTAGE = battery.ErrorCode.ERROR_CODE_HIGH_VOLTAGE
+    """Voltage is too high."""
+
+    LOW_VOLTAGE = battery.ErrorCode.ERROR_CODE_LOW_VOLTAGE
+    """Voltage is too low."""
+
+    HIGH_TEMPERATURE = battery.ErrorCode.ERROR_CODE_HIGH_TEMPERATURE
+    """Temperature is too high."""
+
+    LOW_TEMPERATURE = battery.ErrorCode.ERROR_CODE_LOW_TEMPERATURE
+    """Temperature is too low."""
+
+    HIGH_HUMIDITY = battery.ErrorCode.ERROR_CODE_HIGH_HUMIDITY
+    """Humidity is too high."""
+
+    EXCEEDED_SOP_CHARGE = battery.ErrorCode.ERROR_CODE_EXCEEDED_SOP_CHARGE
+    """Charge current has exceeded component bounds."""
+
+    EXCEEDED_SOP_DISCHARGE = battery.ErrorCode.ERROR_CODE_EXCEEDED_SOP_DISCHARGE
+    """Discharge current has exceeded component bounds."""
+
+    SYSTEM_IMBALANCE = battery.ErrorCode.ERROR_CODE_SYSTEM_IMBALANCE
+    """The battery blocks are not balanced with respect to each other."""
+
+    LOW_SOH = battery.ErrorCode.ERROR_CODE_LOW_SOH
+    """The State of health is low."""
+
+    BLOCK_ERROR = battery.ErrorCode.ERROR_CODE_BLOCK_ERROR
+    """One or more battery blocks have failed."""
+
+    CONTROLLER_ERROR = battery.ErrorCode.ERROR_CODE_CONTROLLER_ERROR
+    """The battery controller has failed."""
+
+    RELAY_ERROR = battery.ErrorCode.ERROR_CODE_RELAY_ERROR
+    """The battery's DC relays have failed."""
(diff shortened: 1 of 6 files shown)


llucax:
I will release v0.4.0 after this is merged.

llucax:
Enabled auto-merge.

src/frequenz/client/microgrid/_component_error.py
```diff
@@ -0,0 +1,198 @@
...
+
+        Args:
+            raw: The protobuf message to get the values from.
+
+        Returns:
+            The new instance with the values from the protobuf message.
+        """
+        return cls(
+            code=BatteryErrorCode.from_pb(raw.code),
+            level=ErrorLevel(raw.level),
```
shsms: I think we shouldn't assume there will only be `WARN` and `CRITICAL` levels, because we don't know what kind of new component types encounter and what kind of error levels they have.  Much safer to have a `from_pb` for `ErrorLevel` too, like the other enum types.

src/frequenz/client/microgrid/_component_error.py
```diff
@@ -0,0 +1,198 @@
...
+
+        Args:
+            raw: The protobuf message to get the values from.
+
+        Returns:
+            The new instance with the values from the protobuf message.
+        """
+        return cls(
+            code=InverterErrorCode.from_pb(raw.code),
+            level=ErrorLevel(raw.level),
```
shsms: here too.

src/frequenz/client/microgrid/_component_states.py
```diff
@@ -97,3 +192,64 @@ def from_pb(cls, evc_state: ev_charger.ComponentState) -> Self:
...
+    """The inverter is in a faulty state."""
+
+    UNAVAILABLE = inverter.ComponentState.COMPONENT_STATE_UNAVAILABLE
+    """The inverter is online, but currently unavailable.
+
+    Possibly due to a pre- scheduled maintenance.
+    """
+
+    UNKNOWN = inverter.ComponentState.COMPONENT_STATE_UNKNOWN
+    """The state is provided by the device, but it is not one of the above states."""
```
shsms: I think we should say component, not device.

```suggestion
    """A state is provided by the component, but it is not one of the above states."""

File: src/frequenz/client/microgrid/_component_states.py [hunk 4]
```diff
@@ -80,6 +80,9 @@ class EVChargerComponentState(Enum):
     INTERRUPTED = ev_charger.ComponentState.COMPONENT_STATE_INTERRUPTED
     """The component is interrupted."""
 
+    UNKNOWN = ev_charger.ComponentState.COMPONENT_STATE_UNKNOWN
+    """The state is provided by the device, but it is not one of the above states."""
```
shsms: I think we should say component, not device.

```suggestion
    """A state is provided by the component, but it is not one of the above states."""

File: src/frequenz/client/microgrid/_component_states.py [hunk 5]
```diff
@@ -6,7 +6,102 @@
...
+    SWITCHING_ON = battery.ComponentState.COMPONENT_STATE_SWITCHING_ON
+    """
+    The battery is starting up and needs some time to become fully operational.
+    """
+
+    SWITCHING_OFF = battery.ComponentState.COMPONENT_STATE_SWITCHING_OFF
+    """The battery is switching off and needs some time to fully shut down."""
+
+    UNKNOWN = battery.ComponentState.COMPONENT_STATE_UNKNOWN
+    """The state is provided by the device, but it is not one of the above states."""
```
shsms: I think we should say component, not device.

```suggestion
    """A state is provided by the component, but it is not one of the above states."""

File: src/frequenz/client/microgrid/_component_states.py [hunk 6]
```diff
@@ -94,6 +94,6 @@ def from_pb(cls, evc_state: ev_charger.ComponentState) -> Self:
             Enum value corresponding to the protobuf message.
         """
         if not any(t.value == evc_state for t in EVChargerComponentState):
-            return cls(cls.UNSPECIFIED)
+            return cls(cls.UNKNOWN)
```
shsms, line 194: I guess we can't escape having to repurpose something.  Now we are giving extra meaning to `UNKNOWN` instead, saying that either the modbus drivers couldn't recognize the state (and it gets set by the api service), or the client couldn't recognize the state.

shsms: We can have `UNKNOWN_TO_CLIENT = -1` and because protobuf enums can't have negative numbers, it will always be unique, but might be overkill,  and hard to name.

shsms: There are some enums that don't have a `UNKNOWN`, and for those, we're still returning UNSPECIFIED when a code is unknown to the client.  So maybe it is worth making them all consistent.

src/frequenz/client/microgrid/_component_states.py
llucax: This comes from the API specs, I guess we could change the wording here but not sure if we should _elevate_ it to the API specs. What do you think?

src/frequenz/client/microgrid/_component_states.py
llucax: Yeah, not sure. If you look at the gRPC errors, some codes can be produced by the server or client, like `INTERNAL` or `RESOURCE_EXHAUSTED` (see https://grpc.github.io/grpc/core/md_doc_statuscodes.html), I think we can see this similarly and say that the estate is unknown, I'm not sure if it is that relevant to the **user of the client** to know if it is unknown to the server or client, I can only see the case where the client is outdated compared to the API specs and the users somehow re-generate its own bindings and use that instead to see if the code is actually known by the server.

IMHO `UNKNOWN` is definitely more accurate here, and if some estate don't have this in the spec, I think that should be a specs bug. @tiyash-basu-frequenz ?

llucax:
To speed up things I accepted the suggestion to improve the wording for the UNKNOWN estates and added a `from_pb` method to `ErrorLevel`. I disabled auto-merge because I need to squash the fixup once I get a LGTM.

llucax:
Actually I'm leaving the auto-merge enabled. Since the fixup commit is not signed-of-by, the automerge won't happen until I squash the commits, so if I get an approval I can squash, force-push and the PR will be queued if the resulting diff doesn't change, so it won't require a second approval after the squashing.

src/frequenz/client/microgrid/_component_states.py
llucax: (ah, Tiyash will probably not answer soon, so we shouldn't wait for his input.

shsms: > IMHO UNKNOWN is definitely more accurate here, and if some estate don't have this in the spec, I think that should be a specs bug.

I think UNKNOWN was introduced for component states where we can have a bug in parsing the raw data.  That's why the the descriptions for all UNKNOWNs say "Component has state, but we understand it".  For other enums where the data doesn't come from hardware, the API doesn't have an UNKNOWN option.

But we are trying to re-purpose UNKNOWN to suggest server-client communication failure, which could happen to all enums, and that's why I think we are trying to give two meanings.

One more option is to say `UNKNOWN = -1` in all python definitions of enums that don't have unknown.

We can keep for later if you prefer, but until then, we'll have UNKNOWN and UNSPECIFIED both being repurposed for something they were not originally meant to do.

shsms:
Fixes look good to me,  can approve once squashed.

shsms (approved):
oops, just saw you wanted early approval, so you could merge asynchronously.

llucax:
You can have another look if you want, but the only that could be messed up is the split of the commits, as long as the resulting diff is the same, the approval still stands.

src/frequenz/client/microgrid/_component_states.py
llucax: Yeah, I think this goes beyond the microgrid API, maybe it would be a good topic for the API meeting.

PR #61, opened 2024-07-01
Configure dependabot to have separate minor updates for in-devel deps

llucax:
The `frequenz-client-base` and `frequenz-microgrid-betterproto` dependencies are in-development and every minor update is potentially (and very likely) a breaking change. Because of this we want dependabot to create separate PRs for minor updates for these dependencies.

For patch updates we still want to group them together, because they not supposed to be breaking changes.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,11 +25,36 @@ updates:
         update-types:
           - "minor"
           - "patch"
+        patterns:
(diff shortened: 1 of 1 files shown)


tiyash-basu-frequenz (approved):

llucax:
Will test if it works out, if it does we can do the same in other projects where it is very annoying that dependabot updates can't be applied because we depend on some 0.x version that gets automatically bumped as a minor update.

PR #70, opened 2024-07-01
Fix patterns in dependabot.yml

llucax:
The new dependabot file updated by #61 had many issues that we fix in this commit:

* The patterns should be glob-like patterns instead of regex patterns.
* We should `exclude-patterns` instead of `patterns` (with a negating regex) to exclude some dependencies.
* We don't actually need the `*-minor` groups, as anything that is not grouped will have a separate PR anyway.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,36 +25,24 @@ updates:
         update-types:
           - "minor"
           - "patch"
-        patterns:
-          - "^(?!frequenz-client-base\\[grpclib\\]).*$"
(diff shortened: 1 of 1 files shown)


llucax:
This time I tested this in my fork and it seems to work correctly now:

* https://github.com/llucax/frequenz-client-microgrid-python/actions/workflows/dependabot/dependabot-updates
* https://github.com/llucax/frequenz-client-microgrid-python/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen

llucax:
I will force-merge as the current config is broken anyway.

PR #81, opened 2024-08-29
Disable` pylint`s `no-name-in-module` check

llucax:
We can disable this rule because `mypy` already checks for this kind of errors and this also fixes a long-standing issue with `pylint` that can't figure out protobuf generate stuff have some members:

* https://github.com/pylint-dev/pylint/issues/6281

And now that we disabled this check we can just use the protobuf modules instead of having to declare one alias per each symbol we need. We did this before just so we only need to use a `pylint` `disable=no-name-in-module` one time in the import instead each time we **used** the symbol.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -146,6 +146,7 @@ disable = [
   "unsubscriptable-object",
   # Checked by mypy
   "no-member",
+  "no-name-in-module",
   # Checked by flake8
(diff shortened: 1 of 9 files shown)


llucax:
Skipping release notes as these are all internal changes.

Marenz (approved):

PR #85, opened 2024-09-02
Disable SSL by default

llucax:
The microgrid API does not use SSL by default, so it makes sense to disable it by default in the client. This was the behavior in previous versions, and it was changed by mistake in a previous commit.

Also prepare the release notes for the next patch release.


llucax:
Tested against the sandbox.

llucax:
I just sneaked in a dependabot fix too. Will force merge as this is all pretty trivial and want to make the release soon so we can release the SDK too.

PR #100, opened 2024-11-18
Update the `client-base` dependency to v0.8.0

llucax:
We now use the new suggested way to get the `stub`, so it has proper async type hints.

We also enable the old behavior to retry on stream exhaustion because we don't expect data streaming to end normally, so we want to keep retrying if that happens.


llucax:
Draft because it needs a release of v0.8.0 (and an update of the dependency in `pyproject.toml`).

Marenz:
fyi, PR in base-client was merged, I updated URL in pyproject file

Marenz (dismissed):

llucax:
I'm sneaking in a commit to stop getting those annoying warnings about `--platform` from docker.

PR #116, opened 2025-02-20
 Widen `frequenz-client-base` dependency to allow `v0.9.0` 

shsms:
This requires that the channels version is at least v1.6.1.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ requires-python = ">= 3.11, < 4"
 dependencies = [
   "frequenz-api-microgrid >= 0.15.3, < 0.16.0",
   "frequenz-channels >= 1.0.0-rc1, < 2.0.0",
-  "frequenz-client-base >= 0.8.0, < 0.9",
+  "frequenz-client-base >= 0.8.0, < 0.10.0",
   "grpcio >= 1.54.2, < 2",
   "protobuf >= 4.21.6, < 6",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -37,8 +37,8 @@ classifiers = [
 requires-python = ">= 3.11, < 4"
 dependencies = [
   "frequenz-api-microgrid >= 0.15.3, < 0.16.0",
-  "frequenz-channels >= 1.0.0-rc1, < 2.0.0",
-  "frequenz-client-base >= 0.8.0, < 0.9",
+  "frequenz-channels >= 1.6.1, < 2.0.0",
+  "frequenz-client-base >= 0.9.0, < 0.10",
```
llucax:
```suggestion
  "frequenz-channels >= 1.0.0-rc1, < 2.0.0",
  "frequenz-client-base >= 0.8.0, < 0.10",
```

llucax (commented):
If this is done for compatibility reasons, you don ´t need to bump any **minimum** dependencies here, just widen the range for `frequenz-client-base`, as no code changes are needed to work with the newer version, we are still compatible with the old version(s).

You don't need to bump the channels version at all, if any other dependency require the new version, the dependency will be resolved to 1.6.1 automatically, as this repo is still compatible with any 1.x.x version.

pyproject.toml
shsms: done

llucax (approved):

PR #122, opened 2025-03-17
Add ID wrapper classes

llucax:
Introduce `MicrogridId` and `ComponentId` classes to replace plain integer IDs. These classes provide type safety and prevent accidental errors by:

- Making it impossible to mix up microgrid and component IDs (equality comparisons between different ID types always return false).
- Preventing accidental math operations on IDs.
- Providing clear string representations for debugging (MID42, CID42).
- Ensuring proper hash behavior in collections.

### DIFF
--- a/src/frequenz/client/microgrid/_id.py
+++ b/src/frequenz/client/microgrid/_id.py
@@ -1,0 +1,102 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+
+"""Strongly typed IDs for microgrids and components."""
+
+
+class MicrogridId:
+    """A unique identifier for a microgrid."""
+
+    def __init__(self, id_: int, /) -> None:
+        """Initialize this instance.
+
+        Args:
+            id_: The numeric unique identifier of the microgrid.
+
+        Raises:
+            ValueError: If the ID is negative.
+        """
+        if id_ < 0:
+            raise ValueError("Microgrid ID can't be negative.")
+        self._id = id_
+
+    def __int__(self) -> int:
+        """Return the numeric ID of this instance."""
+        return self._id
+
+    def __eq__(self, other: object) -> bool:
+        """Check if this instance is equal to another object."""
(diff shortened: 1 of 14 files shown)


llucax:
This is to minimize the changes between the v0.15 and v0.17 clients, so it is easier to maintain a v0.17 branch in the SDK, as this change is very intrusive and could make merging new changes to a branch with this pretty conflict-prone.

src/frequenz/client/microgrid/_id.py
```diff
@@ -0,0 +1,102 @@
...
+    def __int__(self) -> int:
+        """Return the numeric ID of this instance."""
+        return self._id
+
+    def __eq__(self, other: object) -> bool:
+        """Check if this instance is equal to another object."""
+        # This is not an unidiomatic typecheck, that's an odd name for the check.
+        # isinstance() returns True for subclasses, which is not what we want here.
+        # pylint: disable-next=unidiomatic-typecheck
+        return type(other) is MicrogridId and self._id == other._id
```
shsms, line 32: We can still do `isinstance` and disable inheritance for the class using the `typing.final` decorator.

shsms (approved):
Just an optional comment.

src/frequenz/client/microgrid/_id.py
llucax: Maybe, not sure if `typing.final` also performs a runtime check or not, if it doesn't, then the current code is safer as it will be checked at runtime. In practice I don't think it should be a big deal anyway, I doubt anyone would want to subclass these IDs, but in any case adding `typing.final()` might be nice addition to make the intention clearer.

llucax: https://docs.python.org/3.11/library/typing.html#typing.final

> There is no runtime checking of these properties

llucax: * https://github.com/frequenz-floss/frequenz-client-microgrid-python/pull/123

PR #144, opened 2025-05-14
Add sensors listing and data streaming

### LINKED ISSUE #136: Expose sensors
### What's needed?

We need to be able to get sensors and sensor data.

### Proposed solution

Add a new client methods based as much as possible on the API v0.17, so it is easier to migrate:

* `list_sensors()`
* `receive_sensor_data_samples_stream()`

### Additional context

* v0.17 sensors metadata and data samples: https://github.com/frequenz-floss/frequenz-api-common/blob/v0.x.x/proto/frequenz/api/common/v1/microgrid/sensors/sensors.proto
* v0.17 sensors RPC calls: [`ListSensors`](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/2064cf159f3e4d7d32b4f41b8f1f959bb4954634/proto/frequenz/api/microgrid/v1/microgrid.proto#L65-L89), [`ReceiveSensorDataStream`](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/2064cf159f3e4d7d32b4f41b8f1f959bb4954634/proto/frequenz/api/microgrid/v1/microgrid.proto#L125-L131)
* v0.15 sensors component type: https://github.com/frequenz-floss/frequenz-api-common/blob/59dcdc0146de9b64a5f25f55165617237098a343/proto/frequenz/api/common/components.proto#L37-L38
* v0.15 sensors types: https://github.com/frequenz-floss/frequenz-api-common/blob/59dcdc0146de9b64a5f25f55165617237098a343/proto/frequenz/api/common/components.proto#L107-L132
* v0.15 sensor metadata and data: https://github.com/frequenz-floss/frequenz-api-microgrid/blob/v0.15.x/proto/frequenz/api/microgrid/sensor.proto

llucax:
This pull request introduces support for sensors to the microgrid client and includes several bug fixes.

A key focus of this update is **forward-compatibility** with the upcoming microgrid API v0.17. The new sensor-related RPCs (`list_sensors()`, `stream_sensor_data()`) and their associated data structures (`AggregatedMetricValue`, `Lifetime`) have been designed based on the v0.17 specifications. This approach aims to simplify the future migration process for users.

The `list_sensors()` functionality introduces a new sensor class hierarchy, offering a more Pythonic and organized way to interact with different sensor types.

The main changes are:

*   **Sensor Support:**

    *   Addition of `list_sensors()` RPC to retrieve available sensors.
    *   Addition of `stream_sensor_data()` RPC to stream data from sensors.
    *   Introduction of data structures like `SensorId`, `Lifetime`, and `AggregatedMetricValue` to support current and future sensor functionalities.

*   **Bug Fixes:**

    *   Resolved an issue where microgrid metadata would incorrectly default to a (0,0) location if the actual location was missing.
    *   Ensured that broadcasters are properly stopped and cleaned up during client disconnection or exit, preventing potential resource leaks.

*   **Testing:**

    *   Added tests for microgrid metadata retrieval.

Fixes #136.

### DIFF
--- a/src/frequenz/client/microgrid/_client.py
+++ b/src/frequenz/client/microgrid/_client.py
@@ -203,7 +203,7 @@ class MicrogridApiClient(client.BaseApiClient[microgrid_pb2_grpc.MicrogridStub])
             return Metadata()
 
         location: Location | None = None
-        if microgrid_metadata.location:
+        if microgrid_metadata.HasField("location"):
             location = Location(
                 latitude=microgrid_metadata.location.latitude,
                 longitude=microgrid_metadata.location.longitude,
--- a/src/frequenz/client/microgrid/_client.py
+++ b/src/frequenz/client/microgrid/_client.py
@@ -6,17 +6,23 @@
 from __future__ import annotations
 
 import asyncio
+import itertools
 import logging
 from collections.abc import Callable, Iterable, Set
 from dataclasses import replace
-from typing import Any, TypeVar
+from typing import Any, TypeVar, assert_never
 
 from frequenz.api.common import components_pb2, metrics_pb2
-from frequenz.api.microgrid import microgrid_pb2, microgrid_pb2_grpc
+from frequenz.api.microgrid import microgrid_pb2, microgrid_pb2_grpc, sensor_pb2
 from frequenz.channels import Receiver
 from frequenz.client.base import channel, client, retry, streaming
 from google.protobuf.empty_pb2 import Empty
 from typing_extensions import override
+
+from frequenz.client.microgrid._id import SensorId
+from frequenz.client.microgrid.sensor._base import Sensor
+from frequenz.client.microgrid.sensor._data import SensorDataSamples, SensorMetric
+from frequenz.client.microgrid.sensor._data_proto import sensor_data_samples_from_proto
 
 from ._component import (
     Component,
@@ -98,6 +104,12 @@ class MicrogridApiClient(client.BaseApiClient[microgrid_pb2_grpc.MicrogridStub])
         self._broadcasters: dict[
             ComponentId, streaming.GrpcStreamBroadcaster[Any, Any]
         ] = {}
+        self._sensor_data_broadcasters: dict[
+            str,
+            streaming.GrpcStreamBroadcaster[
+                microgrid_pb2.ComponentData, SensorDataSamples
+            ],
+        ] = {}
         self._retry_strategy = retry_strategy
 
     @property
@@ -119,15 +131,22 @@ class MicrogridApiClient(client.BaseApiClient[microgrid_pb2_grpc.MicrogridStub])
         exc_tb: Any | None,
     ) -> bool | None:
         """Close the gRPC channel and stop all broadcasters."""
-        exceptions = [
+        exceptions = list(
             exc
             for exc in await asyncio.gather(
-                *(broadcaster.stop() for broadcaster in self._broadcasters.values()),
+                *(
+                    broadcaster.stop()
+                    for broadcaster in itertools.chain(
+                        self._broadcasters.values(),
+                        self._sensor_data_broadcasters.values(),
+                    )
+                ),
                 return_exceptions=True,
             )
             if isinstance(exc, BaseException)
-        ]
+        )
         self._broadcasters.clear()
+        self._sensor_data_broadcasters.clear()
 
         result = None
         try:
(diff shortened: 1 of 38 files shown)


llucax:
I still need to do some proper testing against a real microgrid API, but I'm pretty confident things shouldn't change much after the tests, so it should be fairly safe to review.

This PR aligns as much as possible with:

* #94

src/frequenz/client/microgrid/_client.py
```diff
@@ -203,7 +203,7 @@ async def metadata(self) -> Metadata:
             return Metadata()
 
         location: Location | None = None
-        if microgrid_metadata.location:
+        if microgrid_metadata.HasField("location"):
```
shsms, line 254: What happened to all our deployments at https://en.wikipedia.org/wiki/Null_Island? 😱

src/frequenz/client/microgrid/_lifetime.py
```diff
@@ -0,0 +1,51 @@
...
+        if self.start is not None and self.end is not None and self.start > self.end:
+            raise ValueError("Start must be before or equal to end.")
+
+    def active_at(self, timestamp: datetime) -> bool:
+        """Check whether this lifetime is active at a specific timestamp."""
+        if self.start is not None and self.start > timestamp:
+            return False
+        if self.end is not None:
+            return self.end >= timestamp
+        # Both are None, so it is always active
```
shsms, line 45: Only `self.end` is known to be `None` with the above checks.  `self.start` could be `None` or earlier than the given timestamp.

tests/test_lifetime.py
```diff
@@ -0,0 +1,322 @@
...
+            start_type=None,
+            end_type=_Time.NOW,
+            expected_active=True,
+        ),
+        _ActivityTestCase(
+            name="now_start-now_end",
+            start_type=_Time.NOW,
+            end_type=_Time.NOW,
+            expected_active=True,
+        ),
```
shsms, line 252: missed case: start=None, end=PAST?  Probably need a _Time.PLUSQUAM

src/frequenz/client/microgrid/sensor/_barometer.py
```diff
@@ -0,0 +1,21 @@
...
+
+from ._base import Sensor
+from ._category import SensorCategory
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class Barometer(Sensor):
+    """A barometer sensor.
+
+    Measures pressure.
```
shsms, line 17:
```suggestion
    Measures atmospheric pressure.

File: src/frequenz/client/microgrid/sensor/_base.py [hunk 5]
```diff
@@ -0,0 +1,81 @@
...
+from functools import cached_property
+from typing import Any, Self
+
+from .._id import SensorId
+from .._lifetime import Lifetime
+from ._category import SensorCategory
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class Sensor:
```
shsms, line 17: Come on @llucax, make more commits!

src/frequenz/client/microgrid/sensor/_base.py
```diff
@@ -0,0 +1,81 @@
...
+    """The category of this sensor.
+
+    Note:
+        This should not be used normally, you should test if a sensor
+        [`isinstance`][] of a concrete sensor class instead.
+
+        It is only provided for using with a newer version of the API where the
+        client doesn't know about a new category yet (i.e. for use with
+        [`UnrecognizedSensor`][frequenz.client.microgrid.sensor.UnrecognizedSensor])
+        and in case some low level code needs to know the category of a sensor.
```
shsms, line 37: This is why it can't be an ABC?  you're blocking `__new__` below anyway.

src/frequenz/client/microgrid/sensor/_base.py
```diff
@@ -0,0 +1,81 @@
...
+        return super().__new__(cls)
+
+    def active_at(self, timestamp: datetime) -> bool:
+        """Check whether this sensor is active at a specific timestamp."""
+        return self.operational_lifetime.active_at(timestamp)
+
+    @cached_property
+    def active(self) -> bool:
+        """Whether this sensor is currently active."""
+        return self.active_at(datetime.now(timezone.utc))
```
shsms, line 66: Not sure you can make this a `cached_property`.  `datetime.now` is not a constant value

src/frequenz/client/microgrid/_lifetime.py
```diff
@@ -0,0 +1,51 @@
...
+        return True
+
+    @cached_property
+    def active(self) -> bool:
+        """Whether this lifetime is currently active."""
+        return self.active_at(datetime.now(timezone.utc))
```
shsms, line 51: This shouldn't be a `cached_property`, because then you can do this with less assumptions.  This property is not expected to be in any hot path, so this is too much optimization anyway.

src/frequenz/client/microgrid/sensor/_category.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@enum.unique
+class SensorCategory(enum.Enum):
+    """The known categories of sensors that can be present in a microgrid."""
+
+    UNSPECIFIED = 0
+    """Unspecified sensor category."""
+
+    THERMOMETER = 1
```
shsms, line 16: These numbers are copied from proto?  If you don't want to import the generated code here and assign from that,  maybe you should add a link to the proto as a comment.

shsms (commented):
Taking a break in the middle of the big commit.

llucax:
Sorry @shsms, of course I was wrong and after doing some tests with the real API I found out the specs for sensors are very old (even in `common` v0.x.x) and will be changed quite a bit in the future (for example the type will be gone, as we have weather stations that provide many metrics, so they don't fall into one specific type). Also the `SensorMetric` might be gone (and we might use the general `Metric`, shared with components), but this is still to be decided, so for now I will keep it.

So this PR will need some big update.

I will split it though, as I accumulated enough unrelated stuff to be worth of a separate PR.

src/frequenz/client/microgrid/_client.py
llucax: We are fine, now microgrids in Null Island need to provide a explicit (0, 0) location :laughing:

src/frequenz/client/microgrid/sensor/_barometer.py
llucax: Most text here is copied from the API, so maybe you can create an issue in the API specs to suggests improvements there too.

src/frequenz/client/microgrid/sensor/_base.py
llucax: I don't see the point on splitting commits that are only adding files, you can just mark files as viewed and that's it. I guess it is very related to your review workflow. But if it makes reviewing easier for you, I can split them, sure, but only if you ask nicely :smile:

src/frequenz/client/microgrid/sensor/_base.py
llucax: I did this a long time ago in #94, and to be honest I'm not sure if I tried to make it a ABC or not and if I had issues if I did. I would have to refresh my mind a bit. But in any case, I will remove all the class hierarchy for sensors, as we are removing the sensor type anyways, so we can leave that for #94.

src/frequenz/client/microgrid/sensor/_base.py
llucax: Yeah, good point, it doesn't make sense for `active`. :+1:  :facepalm:

src/frequenz/client/microgrid/sensor/_category.py
llucax: Yeah, we also talked about this in #94. I'm not a 100% convinced about this, but since these numbers should NEVER EVER change, and we are writing the wrappers manually, I thought I could just skip the generated wrappers, there no value in using them, and if we use them it means all the protobuf definition will be imported when using this simple Python enum (it will probably be imported anyway indirectly at some point, at least if parsing from protobuf is involved, so this is probably more a philosophical issue and a practical one, I know :D). 

But the code also looked really ugly or I had to add exceptions for long lines when using stuff like `sensor_pb2.SENSOR_CATEGORY_THERMOMETER` (with sensors is also not that bad, but for components there are enum names that are really long, and will get worse when they are named `ElectricalComponentCategory.ELECTRICAL_COMPONENT_CATEGORY_INVERTER`).

Maybe we can have a brief meeting to discuss this, I was hoping to open the discussion when I move this to `api-common`, whichever we pick it is a backwards compatible change.

llucax:
Split out the bugfixes to:

* #145

I will probably open a new PR for the sensor stuff, so I leave this PR/branch as is, as the sensors specs seem to be changing, I want to have this PR under my sleeve in case case we decide to go back to sensor types somehow.

src/frequenz/client/microgrid/sensor/_base.py
llucax: I decided to remove these active properties methods, if we need to add them is a non-breaking change, and they make no sense right now, as sensors will always be active in this sense.

src/frequenz/client/microgrid/_lifetime.py
llucax: Fixed.

tests/test_lifetime.py
llucax: 😑 Added.

llucax:
Superseded by:

* #146

PR #146, opened 2025-05-15
Add sensors listing and data streaming

### LINKED ISSUE #136: Expose sensors

llucax:
This pull request introduces support for sensors to the microgrid client.

A key focus of this update is **forward-compatibility** with the upcoming microgrid API v0.17. The new sensor-related RPCs (`list_sensors()`, `stream_sensor_data()`) and their associated data structures have been designed based on the v0.17 specifications. This approach aims to simplify the future migration process for users.

Fixes #136.


llucax:
@shsms I should have addressed all comments in #144. @tiyash-basu-frequenz maybe it would be good if you have a look to see if there is nothing that goes against future directions.

llucax:
OK, it looks like we are going to unify the `SensorMetric` and `Metric` for components and the metric samples too. But I think I would do that in a separate PR once it is more settled, as this is super fresh.

llucax:
BTW, now I tested this in a real microgrid and it seems to be working fine.

llucax:
@shsms I hope you are happier with the commit split now too :pray:

llucax:
Will rename `Lifetime.active` to `Lifetime.is_operational_now()` (no property, since it is an "expensive" operation), and `active_at` -> `is_operational_at()`.

src/frequenz/client/microgrid/_id.py
```diff
@@ -105,3 +105,53 @@ def __repr__(self) -> str:
...
+            raise ValueError("Sensor ID can't be negative.")
+        self._id = id_
+
+    def __int__(self) -> int:
+        """Return the numeric ID of this instance."""
+        return self._id
+
+    def __eq__(self, other: object) -> bool:
+        """Check if this instance is equal to another object."""
+        # This is not an unidiomatic typecheck, that's an odd name for the check.
```
shsms: I feel like whether it is odd or not is out of scope for this comment.

```suggestion
        # This is not an unidiomatic typecheck, pylint is wrong.

File: src/frequenz/client/microgrid/_util.py [hunk 2]
```diff
@@ -0,0 +1,50 @@
...
+        enum_type: The python enum type to convert to,
+            typically an enum class.
+
+    Returns:
+        The resulting python enum value if the protobuf value is known, otherwise
+            the input value converted to a plain `int`.
+    """
+    # A protobuf enum value is a NewType of int, so we make sure we have a pure int at
+    # this point.
+    value = int(value)
```
shsms: NewType is a static-checker-only thing.  At run time, it is actually just the int value, so this cast is not necessary.

From: https://docs.python.org/3/library/typing.html#newtype

> More precisely, the expression some_value is Derived(some_value) is always true at runtime.

tests/test_sensor.py
```diff
@@ -0,0 +1,152 @@
...
+    manufacturer=_SENSOR.manufacturer,
+    model_name=_SENSOR.model_name,
+)
+
+_DIFFERENT_BOTH_ID = Sensor(
+    id=SensorId(2),
+    name=_SENSOR.name,
+    manufacturer=_SENSOR.manufacturer,
+    model_name=_SENSOR.model_name,
+)
```
shsms: is there a typo here?  They both seem to be the same.

shsms (commented):
Need to continue from `Add wrappers for sensor data`.

src/frequenz/client/microgrid/sensor.py
```diff
@@ -48,3 +73,170 @@ def __str__(self) -> str:
...
+
+    INTERNAL = 2
+    """An internal error within the sensor."""
+
+
+@dataclass(frozen=True, kw_only=True)
+class SensorStateSample:
+    """A sample of state, warnings, and errors for a sensor at a specific time."""
+
+    sampled_at: datetime
```
shsms, line 145: wasn't it supposed to be `sample_time`?

src/frequenz/client/microgrid/sensor.py
```diff
@@ -48,3 +73,170 @@ def __str__(self) -> str:
...
+    """The metrics that can be reported by sensors in the microgrid.
+
+    These metrics correspond to various sensor readings primarily related to
+    environmental conditions and physical measurements.
+    """
+
+    UNSPECIFIED = 0
+    """Default value (this should not be normally used and usually indicates an issue)."""
+
+    TEMPERATURE = 1
```
shsms: Time to switch to pb2 names instead of numbers now?

src/frequenz/client/microgrid/_id.py
llucax: Fair, not convinced about the plain **wrong** here, but I'll see if I can put it in a more relevant way. Although this is in every ID class, so I would have to update all comments.

tests/test_sensor.py
llucax: Ah, good catch. No, in the first version I included the microgrid ID too, as in 0.17 it is included, but to get it I needed to do an extra call to `GetMicrogridMetadata` just to get the microgrid ID. Since adding the microgrid ID in the next version is a backwards-compatible change I opted to leave the microgrid ID out, but forgot to remove the case for "both ID". Will remove.

src/frequenz/client/microgrid/sensor.py
llucax: No, actually the current candidate is `snapshot_time` (and the message will be `SensorStateSnapshot`), but it is not changed yet, for stuff that it is already a bit unstable/changing, I just left the name used in 0.17. @tiyash-basu-frequenz do you think this is stable enough to adopt the new names in the Python API already?

src/frequenz/client/microgrid/sensor.py
llucax: I'm still not convinced. This is a pure-python class/interface, all the dealing with protobuf conversion is done in the `_from_proto()` functions. I just happen to pick the same numeric IDs for convenience, to make conversion more easy.

What arguments do you have in favor of using the PB values here?

src/frequenz/client/microgrid/_id.py
shsms: It says it is unidiomatic, you say it is not, so it is wrong.  But more context sounds good, of course.

src/frequenz/client/microgrid/sensor.py
shsms: > This is a pure-python class/interface

We are trying to build a client for the microgrid API.  Not a stand-alone thing.  This is just increasing the chances for bugs, especially for esoteric component types, which we might add support for only later.  To me, this is just overcomplicating things for no apparent benefit.

> What arguments do you have in favor of using the PB values here?

Single source of truth for enums without any manual work.
and just a few days back, I saw some API PRs changing the numbers for enums.

llucax: I understand your point about this being very dependant on the defined protobuf stuff, but all wrappers are protobuf free. I know it is a very theoretical thing but I like it. :D

It is a very minor thing, but the alternative to me is also a very minor thing. The single point of truth is pointless once this is stable, is a write once, never ever touch it again, ever. So I'm not concerned at about about it.

That said, since it is controversial, and the API is currently NOT stable, I can see it might be problematic until we get there, and if we have some issue because of this, I don't want you in my face telling me I told you so, so I'll change it and leave the discussion for a distant future :P

llucax:
Added some new and some fixup! commits to address the comments. One minor advantage of using `int`s is we can add values that are not present in the protobuf yet (but OK, we could still do that only for the non-existing values, but I decided to remove them for now).

src/frequenz/client/microgrid/sensor.py
llucax: Thinking about this, there are **many** things that will change, even if they are considered fairly stable now, I think I prefer to leave the "modernization" of the client interface for a follow-up PR, let's not block this one on that as we are supposed to have people wanting to use this now.

llucax: * https://github.com/frequenz-floss/frequenz-client-microgrid-python/issues/147

src/frequenz/client/microgrid/_id.py
llucax: **THANK YOU**. I had to redo the whole module because of this :stuck_out_tongue_closed_eyes:

shsms:
Will check in the morning.

src/frequenz/client/microgrid/id.py
```diff
@@ -1,157 +1,218 @@
 # License: MIT
 # Copyright © 2025 Frequenz Energy-as-a-Service GmbH
 
-"""Strongly typed IDs for microgrids, components and sensors."""
+r'''Provides strongly-typed unique identifiers for entities.
```
shsms, line 4: This is very nice, also useful outside the microgrid like for the GridpoolID, etc. and also the reporting client.  Maybe this /and/ the subclasses should go to the base client.

shsms (approved):

src/frequenz/client/microgrid/sensor.py
shsms: > I told you so

Oh, I really enjoy doing it, too bad.  But there will be other opportunities. :D

src/frequenz/client/microgrid/id.py
llucax: Yes, next step is move lots of stuff in here to `client-common` and `client-base`.

PR #148, opened 2025-05-22
Fix sensor data streaming without metrics filter

llucax:
When no metric filter was passed to `stream_sensor_data()` the transform callback failed because the `metrics` argument wasn't passed when calling `sensor_data_samples_from_proto()` internally, making the gRPC streaming task stop with an unhandled exception.

This PR makes the `metrics` argument optional for `sensor_data_samples_from_proto()` to fix that.

A test case with no metric filters is added too, and the test case testing metric filters is improved to actually test the filtering.


llucax:
Skipping release notes as this fixes unreleased code.

tiyash-basu-frequenz (approved):
LGTM

PR #151, opened 2025-05-28
Allow base-client to be updated to 0.11.0

Marenz:
In anticipation of the upcoming release of the base-client


hannah-stevenson-frequenz (approved):
LGTM FWIW NTIKA

llucax:
Did you actually tried to use it with the current `client-base` HEAD, because there are some breaking changes that could affect this client, I'm not sure if blindly bumping the dependency is enough to allow for unreleased, untested code :grimacing:

Marenz:
I will find out very soon :laughing:

PR #155, opened 2025-06-02
Update base-client min version to v0.11

Marenz:


RELEASE_NOTES.md
```diff
@@ -6,11 +6,11 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* When receiving streaming data for components, you now need to handle the receiving of the types `StreamStarted`, `StreamRetrying`, `StreamFatalError`.
```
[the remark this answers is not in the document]
llucax: There is no `-` in this file :roll_eyes:

llucax: Maybe add a tip on how to easily get the old behavior back?

```suggestion
* When receiving streaming data for components, you now need to handle the receiving of the types `StreamStarted`, `StreamRetrying`, `StreamFatalError`.

    If you don't care about the new events and just want the old behavior your can always use `Receiver.filter()` to ignore them, for example: `client.meter_data().filter(lambda m: isinstance(m, MeterData))`.
```

llucax (approved):

llucax (changes requested):
No, wait! You need to also update all the type hints for the client streaming methods, they are only returning `XxxData` now, but they should return also `StreamingEvents` or whatever it was called.

RELEASE_NOTES.md
Marenz: I was thinking what kind of usage code example I could give out, this is better than nothing :+1:

Marenz:
Updated. CI will fail until we release base-client v0.11

llucax (changes requested):
Still missing all the component-specific methods using `_new_component_data_receiver()`, like `meter_data()`, `inverter_data()`, etc.

Marenz:
It's a bit inconsistent that there is one method `stream_` that streams data and a bunch of mixed others that might or might not stream something.

llucax:
> It's a bit inconsistent that there is one method stream_ that streams data and a bunch of mixed others that might or might not stream something.

The `xxx_data()` methods are old, in the v0.17 branch all will have the same naming pattern (but we'll also have only one method for all component types, so it won't matter so much). The sensor support was added recently so I wanted to use a more "modern" interface, so people that start using it have a simpler upgrade path.

llucax (dismissed):

llucax:
I'm changing this to a draft until we decide what we do with the new events, if we really want to mix them with real data.

llucax:
I'm thinking that another alternative if we want to move forward more quickly (in case this is blocking something), we could just filter the events here in the client for now, and keep sending only data via the streams. So basically we contain the breaking change in the `client-base` for now until we find a solution we are happy with.

llucax:
But this will quickly spread though other clients, so we need to find a sustainable solution for the `base-client` soon, we shouldn't end up with all clients filtering the events internally (except for dispatch).

PR #165, opened 2025-07-02
Implement `ListComponents`

llucax:
Implements the new `list_components` call in the `MicrogridApiClient`, adds detailed tests for component classes and protobuf conversions, and enhances the `Lifetime` validation message.

- Introduced `list_components` method with component and category filters in the client
- Added many component constructor and proto‐conversion tests, including new `test_list_components`
- Updated `Lifetime` validation to include values in the error message and adjusted tests accordingly

Part of #55.

### DIFF
--- a/src/frequenz/client/microgrid/component/_battery.py
+++ b/src/frequenz/client/microgrid/component/_battery.py
@@ -1,0 +1,129 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Battery component."""
+
+import dataclasses
+import enum
+from typing import Any, Literal, Self, TypeAlias
+
+from frequenz.api.common.v1.microgrid.components import battery_pb2
+
+from ._category import ComponentCategory
+from ._component import Component
+
+
+@enum.unique
+class BatteryType(enum.Enum):
+    """The known types of batteries."""
+
+    UNSPECIFIED = battery_pb2.BATTERY_TYPE_UNSPECIFIED
+    """The battery type is unspecified."""
+
+    LI_ION = battery_pb2.BATTERY_TYPE_LI_ION
+    """Lithium-ion (Li-ion) battery."""
+
+    NA_ION = battery_pb2.BATTERY_TYPE_NA_ION
+    """Sodium-ion (Na-ion) battery."""
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class Battery(Component):
+    """An abstract battery component."""
+
+    category: Literal[ComponentCategory.BATTERY] = ComponentCategory.BATTERY
+    """The category of this component.
+
+    Note:
+        This should not be used normally, you should test if a component
+        [`isinstance`][] of a concrete component class instead.
+
+        It is only provided for using with a newer version of the API where the client
+        doesn't know about a new category yet (i.e. for use with
+        [`UnrecognizedComponent`][frequenz.client.microgrid.component.UnrecognizedComponent])
+        and in case some low level code needs to know the category of a component.
+    """
+
+    type: BatteryType | int
+    """The type of this battery.
+
+    Note:
+        This should not be used normally, you should test if a battery
+        [`isinstance`][] of a concrete battery class instead.
+
+        It is only provided for using with a newer version of the API where the client
+        doesn't know about the new battery type yet (i.e. for use with
+        [`UnrecognizedBattery`][frequenz.client.microgrid.component.UnrecognizedBattery]).
+    """
+
+    # pylint: disable-next=unused-argument
+    def __new__(cls, *args: Any, **kwargs: Any) -> Self:
+        """Prevent instantiation of this class."""
+        if cls is Battery:
+            raise TypeError(f"Cannot instantiate {cls.__name__} directly")
+        return super().__new__(cls)
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class UnspecifiedBattery(Battery):
+    """A battery of a unspecified type."""
+
+    type: Literal[BatteryType.UNSPECIFIED] = BatteryType.UNSPECIFIED
+    """The type of this battery.
+
+    Note:
+        This should not be used normally, you should test if a battery
+        [`isinstance`][] of a concrete battery class instead.
+
+        It is only provided for using with a newer version of the API where the client
+        doesn't know about the new battery type yet (i.e. for use with
+        [`UnrecognizedBattery`][frequenz.client.microgrid.component.UnrecognizedBattery]).
+    """
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class LiIonBattery(Battery):
+    """A Li-ion battery."""
+
+    type: Literal[BatteryType.LI_ION] = BatteryType.LI_ION
+    """The type of this battery.
+
+    Note:
+        This should not be used normally, you should test if a battery
+        [`isinstance`][] of a concrete battery class instead.
+
+        It is only provided for using with a newer version of the API where the client
+        doesn't know about the new battery type yet (i.e. for use with
+        [`UnrecognizedBattery`][frequenz.client.microgrid.component.UnrecognizedBattery]).
+    """
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class NaIonBattery(Battery):
+    """A Na-ion battery."""
+
+    type: Literal[BatteryType.NA_ION] = BatteryType.NA_ION
+    """The type of this battery.
+
+    Note:
+        This should not be used normally, you should test if a battery
+        [`isinstance`][] of a concrete battery class instead.
+
+        It is only provided for using with a newer version of the API where the client
+        doesn't know about the new battery type yet (i.e. for use with
+        [`UnrecognizedBattery`][frequenz.client.microgrid.component.UnrecognizedBattery]).
+    """
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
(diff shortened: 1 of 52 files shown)


llucax:
This should be the last big PR.

src/frequenz/client/microgrid/component/_electrolyzer.py
```diff
@@ -0,0 +1,18 @@
...
+import dataclasses
+from typing import Literal
+
+from ._category import ComponentCategory
+from ._component import Component
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class Electrolyzer(Component):
+    """A electrolyzer component."""
```
shsms:
```suggestion
    """An electrolyzer component."""

File: src/frequenz/client/microgrid/component/_fuse.py [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+class Fuse(Component):
+    """A fuse component.
+
+    Fuses are used to protect components from overcurrents.
+    """
+
+    category: Literal[ComponentCategory.FUSE] = ComponentCategory.FUSE
+    """The category of this component."""
+
+    rated_current: int
```
shsms, line 23: why not use `quantities.Current` here already?

src/frequenz/client/microgrid/component/_voltage_transformer.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Voltage transformers are used to step up or step down the voltage, keeping
+    the power somewhat constant by increasing or decreasing the current.
+
+    If voltage is stepped up, current is stepped down, and vice versa.
+
+    Note:
+        Voltage transformers have efficiency losses, so the output power is always less
+        than the input power.
+    """
```
shsms, line 25: This is more accurately described as a power transformer or simply a transformer.

A voltage transformer is a different thing used to lower voltage levels so that they can be measured: https://en.wikipedia.org/wiki/Voltage_transformer

I already mentioned the naming mismatch to @tiyash-basu-frequenz.

src/frequenz/client/microgrid/component/_battery.py
```diff
@@ -0,0 +1,129 @@
...
+class UnrecognizedBattery(Battery):
+    """A battery of an unrecognized type."""
+
+    type: int
+    """The unrecognized type of this battery."""
+
+
+BatteryTypes: TypeAlias = (
+    LiIonBattery | NaIonBattery | UnrecognizedBattery | UnspecifiedBattery
+)
```
shsms, line 128: Really not a fan of calling these `BatteryTypes`.  Maybe call `class Battery(Component)` as `class AbstractBattery` and `BatteryTypes` -> `Battery`?

src/frequenz/client/microgrid/component/_ev_charger.py
```diff
@@ -0,0 +1,153 @@
...
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class UnrecognizedEvCharger(EvCharger):
+    """An EV charger of an unrecognized type."""
+
+    type: int
+    """The unrecognized type of this EV charger."""
+
+
+EvChargerTypes: TypeAlias = (
```
shsms, line 146: Same

src/frequenz/client/microgrid/component/_inverter.py
```diff
@@ -0,0 +1,153 @@
...
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class UnrecognizedInverter(Inverter):
+    """An inverter component."""
+
+    type: int
+    """The unrecognized type of this inverter."""
+
+
+InverterTypes: TypeAlias = (
```
shsms, line 146: Same

src/frequenz/client/microgrid/component/_types.py
```diff
@@ -0,0 +1,65 @@
...
+    | UnrecognizedEvCharger
+    | UnrecognizedInverter
+)
+
+ProblematicComponentTypes: TypeAlias = (
+    MismatchedCategoryComponent | UnrecognizedComponentTypes | UnspecifiedComponentTypes
+)
+"""All possible component types that has a problem."""
+
+ComponentTypes: TypeAlias = (
```
shsms, line 48: Maybe rename `Component` -> `AbstractComponent` and this to `Component`?

These things are not lists or iterables, so they can't have plural names.

src/frequenz/client/microgrid/component/_component_proto.py
```diff
@@ -0,0 +1,562 @@
...
+            always_print_fields_with_no_presence=True,
+        )
+
+    category_mismatched = False
+    if (
+        metadata_category
+        and isinstance(category, ComponentCategory)
+        and category.name.lower() != metadata_category
+    ):
+        major_issues.append("category_type.metadata does not match the category_type")
```
shsms, line 187: maybe also add the metadata category found here, because that info is not easily available from elsewhere?

src/frequenz/client/microgrid/component/_fuse.py
llucax: So far the client doesn't depend on `quantities` and use plain `float`s (in the previous PRs we also didn't use `Power` for the set power calls). I thought about adding `quantities`, but if we do, I would also use quantities for streaming data, and that can be more challenging because we'll need a mapping of `Metric` -> `Quantity`.

All of this adds also overhead, as creating new instances of non built-in types in Python is very slow compared to working with plain `float`s, and maybe it is better to keep the client simple and performant as it should be mostly used only for low-level stuff.

That said, I think we can discuss adding `quantities`, but I definitely do it as a separate step, not as part of the v0.17 port.

src/frequenz/client/microgrid/component/_voltage_transformer.py
llucax: * https://github.com/frequenz-floss/frequenz-api-common/issues/387

src/frequenz/client/microgrid/component/_battery.py
llucax: > Really not a fan of calling these BatteryTypes

Why?

The rationale here is `Battery` (as a abstract base class) will probably be used in user code much more than `BatteryTypes` (the type alias). Both `isinstance()` and `match` will use the abstract base class, while using `BatteryTypes` is more niche, I think I'm using it only as return types to allow exhaustion checks of the returned value, like:

```py
bat = battery_from_proto(...)  # Not sure if this function exists, just an example :D
match bat:
    case LiIonBattery():
        ...
    ...
    case unexpected:
        assert_never(unexpected)
```

If `battery_from_proto()` returns `Battery` (abstract class) then we can't have exhaustion checks. But I guess it will be less likely that users write functions returning batteries for example.

All that said, I agree `BatteryTypes` can be confusing, in particular because we have the `BatteryType` enum. Not sure what would be a good alternative though, maybe `BatteriesTypeAlias`? Not sure really, nothing I can come up with sounds particularly sexy.

src/frequenz/client/microgrid/component/_ev_charger.py
llucax: Resolving this to keep the discussion in one thread, but whatever solution we agree on will be done for all `*Types` aliases.

src/frequenz/client/microgrid/component/_inverter.py
llucax: Resolving this to keep the discussion in one thread, but whatever solution we agree on will be done for all `*Types` aliases.

src/frequenz/client/microgrid/component/_types.py
llucax: Resolving this to keep the discussion in one thread, but whatever solution we agree on will be done for all `*Types` aliases.

src/frequenz/client/microgrid/component/_battery.py
llucax: Just for completion, if we use `AbstractBattery` we'll end up with:

```py
for component in await client.list_components():
    match component:
        case AbstractBattery():
            ...
    ...
```

Which seems much uglier than the current:
```py
for component in await client.list_components():
    match component:
        case Battery():
            ...
    ...
```

src/frequenz/client/microgrid/component/_component_proto.py
llucax: When issues are found, the original protobuf message will be printed in full, this is why I make issue strings very short.

llucax: Example:

```py
>>> component_from_proto(c)
WARNING:frequenz.client.microgrid.component._component_proto:Found issues in component: category_type.metadata does not match the category_type | Protobuf message:
id: 1
name: "The COMPONENT"
category: COMPONENT_CATEGORY_GRID
category_type {
  inverter {
    type: INVERTER_TYPE_HYBRID
  }
}
```

llucax:
I replied to all comments and applied the suggested fix to the electrolyzer docstring. I will squash to the commit introducing that docstring when the PR is approved.

src/frequenz/client/microgrid/component/_fuse.py
llucax: BTW, we should use quantities too in `add_component_bounds()`, so we would need the `Metric` -> `Quantity` mapping there too.

shsms (approved):

src/frequenz/client/microgrid/component/_fuse.py
llucax: * https://github.com/frequenz-floss/frequenz-client-microgrid-python/issues/166

PR #167, opened 2025-07-04
Implement `ListConnections`

llucax:
Part of #55.

### DIFF
--- a/src/frequenz/client/microgrid/_client.py
+++ b/src/frequenz/client/microgrid/_client.py
@@ -28,6 +28,8 @@ from ._microgrid_info_proto import microgrid_info_from_proto
 from .component._category import ComponentCategory
 from .component._component import Component
 from .component._component_proto import component_from_proto
+from .component._connection import ComponentConnection
+from .component._connection_proto import component_connection_from_proto
 from .component._types import ComponentTypes
 from .metrics._bounds import Bounds
 from .metrics._metric import Metric
@@ -220,6 +222,66 @@ class MicrogridApiClient(client.BaseApiClient[microgrid_pb2_grpc.MicrogridStub])
         )
 
         return map(component_from_proto, component_list.components)
+
+    async def list_connections(  # noqa: DOC502 (raises ApiClientError indirectly)
+        self,
+        *,
+        sources: Iterable[ComponentId | Component] = (),
+        destinations: Iterable[ComponentId | Component] = (),
+    ) -> Iterable[ComponentConnection]:
+        """Fetch all the connections present in the local microgrid.
+
+        Electrical components are a part of a microgrid's electrical infrastructure
+        are can be connected to each other to form an electrical circuit, which can
+        then be represented as a graph.
+
+        The direction of a connection is always away from the grid endpoint, i.e.
+        aligned with the direction of positive current according to the passive sign
+        convention: https://en.wikipedia.org/wiki/Passive_sign_convention
+
+        The request may be filtered by `source`/`destination` component(s) of individual
+        connections.  If provided, the `sources` and `destinations` filters have an
+        `AND` relationship between each other, meaning that they are applied serially,
+        but an `OR` relationship with other elements in the same list.
+
+        Example:
+            If `sources = {1, 2, 3}`, and `destinations = {4,
+            5, 6}`, then the result should have all the connections where:
+
+            * Each `source` component ID is either `1`, `2`, OR `3`; **AND**
+            * Each `destination` component ID is either `4`, `5`, OR `6`.
+
+        Args:
+            sources: The component from which the connections originate.
+            destinations: The component at which the connections terminate.
+
+        Returns:
+            Iterator whose elements are all the connections in the local microgrid.
+
+        Raises:
+            ApiClientError: If the are any errors communicating with the Microgrid API,
+                most likely a subclass of
+                [GrpcError][frequenz.client.microgrid.GrpcError].
+        """
+        connection_list = await client.call_stub_method(
+            self,
+            lambda: self.stub.ListConnections(
+                microgrid_pb2.ListConnectionsRequest(
+                    starts=map(_get_component_id, sources),
+                    ends=map(_get_component_id, destinations),
+                ),
+                timeout=DEFAULT_GRPC_CALL_TIMEOUT,
+            ),
+            method_name="ListConnections",
+        )
+
+        return (
+            conn
+            for conn in map(
+                component_connection_from_proto, connection_list.connections
+            )
+            if conn is not None
+        )
 
     async def set_component_power_active(  # noqa: DOC502 (raises ApiClientError indirectly)
         self,
--- a/src/frequenz/client/microgrid/component/_connection.py
+++ b/src/frequenz/client/microgrid/component/_connection.py
@@ -1,0 +1,72 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+
+"""Component connection."""
+
+import dataclasses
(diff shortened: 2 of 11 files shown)


src/frequenz/client/microgrid/_client.py
```diff
@@ -221,6 +223,66 @@ async def list_components(  # noqa: DOC502 (raises ApiClientError indirectly)
...
+            ApiClientError: If the are any errors communicating with the Microgrid API,
+                most likely a subclass of
+                [GrpcError][frequenz.client.microgrid.GrpcError].
+        """
+        connection_list = await client.call_stub_method(
+            self,
+            lambda: self.stub.ListConnections(
+                microgrid_pb2.ListConnectionsRequest(
+                    starts=map(_get_component_id, sources),
+                    ends=map(_get_component_id, destinations),
```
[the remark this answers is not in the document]
llucax: Nah, protobuf messages accept iterables.

src/frequenz/client/microgrid/component/_connection.py
```diff
@@ -0,0 +1,72 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
```
Marenz: optional, 2025

src/frequenz/client/microgrid/component/_connection.py
```diff
@@ -0,0 +1,72 @@
...
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class ComponentConnection:
+    """A single electrical link between two components within a microgrid.
+
+    A component connection represents the physical wiring as viewed from the grid
+    connection point, if one exists, or from the islanding point, in case of an islanded
+    microgrids.
+
+    Note: Physical Representation
+        This message is not about data flow but rather about the physical
```
Marenz: message? What message is it referring?

src/frequenz/client/microgrid/component/_connection.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Note: Direction
+        The direction of the connection follows the flow of current away from the
+        grid connection point, or in case of islands, away from the islanding
+        point. This direction is aligned with positive current according to the
+        [Passive Sign Convention]
+        (https://en.wikipedia.org/wiki/Passive_sign_convention).
+
+    Note: Historical Data
+        The timestamps of when a connection was created and terminated allows for
```
Marenz: allow*s* -> allow

src/frequenz/client/microgrid/component/_connection.py
```diff
@@ -0,0 +1,72 @@
...
+
+from .._lifetime import Lifetime
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class ComponentConnection:
+    """A single electrical link between two components within a microgrid.
+
+    A component connection represents the physical wiring as viewed from the grid
+    connection point, if one exists, or from the islanding point, in case of an islanded
```
Marenz, line 19: Is "islanding point" a fixed term we use? "island point" would seem more natural to me

Marenz: > the grid connection point, if one exists, or from the islanding point, in case of an islanded microgrids

Also, I wonder if it would be easier to read if we describe this only once, but we repeat this exact phrase at least two more times..

Marenz: Like, the other times we could at least say "grid connection/islanding point" instead of the whole phrase. But maybe there is also a more neutral term

src/frequenz/client/microgrid/component/_connection.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+
+    destination: ComponentId
+    """The unique ID of the component where the connection terminates.
+
+    This is the component towards which the current flows.
+    """
+
+    operational_lifetime: Lifetime = dataclasses.field(default_factory=Lifetime)
+    """The operational lifetime of the connection."""
```
Marenz, line 55: Whats an operational Lifetime?

src/frequenz/client/microgrid/component/_connection.py
llucax: I wouldn't be surprised if I wrote this originally in 2024 :laughing: 

(will update if I need to make other changes to the file)

src/frequenz/client/microgrid/component/_connection.py
llucax: Protobuf message, but yeah, point taken, the wording should be updated to Python lingo (i.e. *class*).

src/frequenz/client/microgrid/component/_connection.py
llucax: Copy&paste from the common API, I can fix it here but we should fix it there too.

src/frequenz/client/microgrid/component/_connection_proto.py
```diff
@@ -0,0 +1,104 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
```
Marenz: 25

src/frequenz/client/microgrid/component/_connection.py
llucax: All this should go to the common API, it is all taken from there. I'm not sure how are we going to cope with keeping these in sync, but I guess this is only an issue until we reach 1.0.0, then changes should be very minimal. Maybe when 1.0.0 is released we need to go through all the docs and check they are consistent.

src/frequenz/client/microgrid/component/_connection.py
llucax: It is still a bit *uncertain*, sort of. For the microgrid API, as long as we don't add a way to get historical data for the component graph, this should probably always be empty. I think it is mostly for historical data to signal what was the timespan this component was actually part of the component graph. Not sure if this relates to the operational mode (like if it was disabled if it counts as operational or not).

This still probably needs quite a bit of though and polishing, I don't think we'll solve it for the time being, but I could add some warning that this field utility is still a bit uncertain. I considered also not parsing this field from the protobuf message and skip it completely here, that would be another option, but we already merged some code that has `operational_lifetime` in it, so it would mean more work to remove it now.

@tiyash-basu-frequenz any recommendations or anything to add?

src/frequenz/client/microgrid/component/_connection_proto.py
llucax: Fixed

src/frequenz/client/microgrid/component/_connection.py
llucax: Fixed

src/frequenz/client/microgrid/component/_connection.py
llucax: Fixed

src/frequenz/client/microgrid/component/_connection.py
llucax: Fixed

src/frequenz/client/microgrid/component/_connection.py
llucax: This one I will pass, as it diverges too much from the source, I'd rather fix it in the common API first to agree in what's the best wording first with @tiyash-basu-frequenz.

src/frequenz/client/microgrid/component/_connection.py
llucax: For now I'm also leaving this one at it is until there is a clear path on how to move forward.

Marenz (approved):

PR #190, opened 2025-09-26
Upgrade the client to v0.18.x (v0.8.0 common)

llucax:
This PR does a minimal upgrade to be able to use the microgrid API v0.18.0 and common API v0.8.0.

It mainly updates the import and perform simple symbol renames, as well as adapting some structural changes. The interface is not affected, changes should be backwards compatible, although some minor behavior changes might be present (for example responses of `set_component_power_xxx()`).

The structural changes are:

* `set_component_power_active()`/`reactive()` changed from unary-unary to unary-stream. For now, we are only receiving the first response and returning that, but this will be improved in the future.

* `add_component_bounds()` changed the Validity enum to have int values to cope with the more flexible API. In the future we'll add the possibility to pass an arbitrary validity.

* `ComponentStateSample`'s errors and warnings are now wrapped in a message with multiple fields, for now we are only retrieving the code. In the future we'll wrap the new fields too.

* `MetricSample.connection` is now also wrapped in a message with multiple fields. We are only getting the connection name for now, but we'll add the new information in the future too.

* Some component categories were renamed, we added the new names but kept the old ones as deprecated. We needed to add the new names in this commit to avoid having to adapt the code too much, as the category name is tied to the `category_specific_info.kind` value, so we needed the new values. We need to disable the "unique" constraint for `ComponentCategory` temporarily to allow for the deprecated aliases.

Other changes (and there are many pending) are left as future improvements. Some of these pending changes are:

* Rename component -> electrical component (including IDs and fields)
* Rename grid -> grid connection point
* Rename voltage transformer -> power transformer
* Rename relay -> breaker
* Rename SOLAR -> PV
* Rename source -> connection
* Add new component categories, Metrics and other new enum fields
* Deprecate UNDERVOLTAGE_SHUTDOWN and old Metric names
* Rename category_speific_metadata -> category_specific_info
* Rename sampled_at -> origin_time / sample_time


Marenz (approved):

tiyash-basu-frequenz (approved):
LGTM! Thanks for this :)

llucax:
I ran some basic tests and it seems it is working as expected, so I will merge.

PR #266, opened 2026-04-27
Prepare release

simonvoelcker:
This should be **v0.18.3**.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,13 +37,13 @@ classifiers = [
 requires-python = ">= 3.11, < 4"
 dependencies = [
   "frequenz-api-common >= 0.8.2, < 1.0.0",
-  "frequenz-api-microgrid >= 0.18.0, < 0.19.0",
+  "frequenz-api-microgrid >= 0.18.0, < 1",
   "frequenz-channels >= 1.6.1, < 2.0.0",
   "frequenz-client-base >= 0.10.0, < 0.12.0",
   "frequenz-client-common >= 0.3.6, < 0.4.0",
   "frequenz-core >= 1.3.0, < 2.0.0",
-  "grpcio >= 1.72.1, < 2",
-  "protobuf >= 6.32.1, < 8",
+  "grpcio >= 1.80.0, < 2",
+  "protobuf >= 6.33.6, < 8",
   "typing-extensions >= 4.13.0, < 5",
 ]
 dynamic = ["version"]
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,17 +1,5 @@
 # Frequenz Microgrid API Client Release Notes
 
-## Summary
-
-<!-- Here goes a general summary of what this release is about -->
-
-## Upgrading
-
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
-
 ## New Features
 
 A new component type, SteamBoiler, was added. Also added a new component category for it.
-
-## Bug Fixes
-
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->


simonvoelcker:
@llucax @tiyash-basu-frequenz anybody?

tiyash-basu-frequenz (dismissed):
LGTM.
Note that v0.8.4 of `frequenz-api-common` is now out, and I will soon release v0.19.0 of `frequenz-api-microgrid`.

simonvoelcker:
@tiyash-basu-frequenz upon trying to merge this last week, the pipeline failed (`ci_checks_max` job) because of an unexpected breaking change in `frequenz-api-microgrid` version `0.18.1`.

I guess it's better to adjust the code to the new version rather than pinning, so I'll look into that :eyes:

simonvoelcker:
Fixed the dependency issues - Please let me know whether it's okay to amend this change to this PR or whether it should go into a separate one.

pyproject.toml
```diff
@@ -36,14 +36,14 @@ classifiers = [
 ]
 requires-python = ">= 3.11, < 4"
 dependencies = [
-  "frequenz-api-common >= 0.8.2, < 1.0.0",
-  "frequenz-api-microgrid >= 0.18.0, < 0.19.0",
+  "frequenz-api-common >= 0.8.4, < 1.0.0",
+  "frequenz-api-microgrid >= 0.19.0, < 0.20.0",
```
llucax: We shouldn't bump yet to 0.19 IMHO. This change was shipped in common 0.8.2, there is no need to bump to 0.19, which is a breaking release, so other code will need to be updated.
```suggestion
  "frequenz-api-common >= 0.8.2, < 1.0.0",
  "frequenz-api-microgrid >= 0.18.0, < 0.19.0",
```

llucax:
> @tiyash-basu-frequenz upon trying to merge this last week, the pipeline failed (`ci_checks_max` job) because of an unexpected breaking change in `frequenz-api-microgrid` version `0.18.1`.

What was the breakage exactly? Also I don't see a v0.18.1 in api-microgrid. Something funny's going on!

llucax:
And after this is merged, the release should be frequenz-client-microgrid v0.18.3, not v0.19.1.

simonvoelcker:
@llucax I thought it should be `0.19.1` because there is a `0.19.0` tag already. Thanks for letting me know, I'll update it in the description.

simonvoelcker:
> What was the breakage exactly?

@llucax right [here](https://github.com/frequenz-floss/frequenz-client-microgrid-python/actions/runs/25113862068/job/73595183459).

This is what I assume happened in `frequenz-api-microgrid`:
1. `v0.18.1` was released but was found to be breaking
2. `v0.18.1` was removed
3. Same code was re-released as `v0.19.0`.

So, to resolve the errors I bumped it along with everything else that was causing issues and now the pipeline passes.

llucax:
> @llucax I thought it should be `0.19.1` because there is a `0.19.0` tag already. Thanks for letting me know, I'll update it in the description.

Yea, it was a failed release.

> This release was supposed to be v0.18.1, please don't use it. Use v0.18.1 instead.

We have one once in a while 😁

pyproject.toml
llucax: Thinking more about it, I'm thinking that if we are saying all v0.x.x releases are compatible now for API repos, because we ship breaking changes as different namespaces in the protobuf side (we create a new `v0alphaX` namespace for every breaking release), then we probably can just use v0.19.x in the client, and also use the range `>= 0.18.x, < 1` as we do now with `api-common` in `api-microgrid`. We can depend independently on a common version if we need new common versions.

I just tried this and it works:

```suggestion
  "frequenz-api-common >= 0.8.2, < 1.0.0",
  "frequenz-api-microgrid >= 0.18.0, < 1",
```

We still release this (client) as v0.18.x, because it is backwards compatible, and we are not really using the v0.19 API yet.

FYI @tiyash-basu-frequenz

simonvoelcker: Made the change :+1:

llucax (approved):

